PR #27, opened 2024-10-25
Closes #26 - Add Health Check for Adapter

### LINKED ISSUE #26: Health Check for Adapter
Check Health of Adapter and the used services using Spring Boot Actuator. The calls to the services are listed below:

1. Outbox: Get Number of Failed Events
- Return Code 200 → Status “UP
- Other Return Codes → Status “DOWN”

2. Kadai: Get Schema Version
- If there is anything returned → Status “UP” (can be empty)
- If Exception returned → Status “DOWN”

3. Camunda REST: Still open, should be a call to the DB

jamesrdi:
### Thanks for your PR! Please fill out the following list :)

---

- [ ] I put the ticket or multiple tickets in review
- [x] Commit message format → Closes #&lt;Issue Number&gt; - Your commit message.
- [ ] Sonarcloud link : \<add the link here>
- [ ] No documentation update needed
- [x] Link to PR with documentation update: [\<add the link here>](https://github.com/kadai-io/kadai-doc/pull/27)
- [ ] No Release Notes needed
- [x] Release Notes :

```
Add the following properties to be able to see the health endpoint:

management.endpoints.web.exposure.include= *
management.endpoint.health.show-details= always
management.health.external-services.include=external-services
management.health.external-services.enabled=true
camundaOutboxService.port= <port where camunda & outbox runs>
camundaOutboxService.address=<address where camunda & outbox runs>

The health check can be found under the endpoint /actuator/health/external-services
```


kadai-adapter-camunda-spring-boot-test/pom.xml
```diff
@@ -137,6 +137,10 @@
             <version>${version.jakarta.annotation}</version>
             <scope>test</scope>
         </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-webflux</artifactId>
```
CRoberto1926: Do we really want to add this dependency, just to use WebTestClient? Can't we use TestRestTemplate or (if it exists) the test version of RestClient?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/HealthCheckTest.java
```diff
@@ -0,0 +1,287 @@
...
+            restTemplate.exchange(
+                Mockito.eq("http://localhost:8081/example-context-root/engine-rest/engine"),
+                Mockito.eq(HttpMethod.GET),
+                Mockito.any(),
+                Mockito.eq(responseType)))
+        .thenReturn(ResponseEntity.ok(engines));
+
+    webTestClient
+        .get()
+        .uri("/actuator/health/externalServices")
```
CRoberto1926: Can we modify the URL to be "/actuator/health/external-services"? (kebap-case instead of camelCase)

kadai-adapter/pom.xml
```diff
@@ -55,6 +55,11 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-tx</artifactId>
         </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-actuator</artifactId>
+        <version>3.3.4</version>
```
CRoberto1926: You can delete this version here because it inherits from spring-boot-dependencies

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+import org.springframework.core.env.Environment;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+
+@Component("externalServices")
```
CRoberto1926: Why didn't you call this `ExternalServicesHealthIndicator`?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+public class HealthCheck implements HealthIndicator {
```
CRoberto1926: Why not `public class ExternalServicesHealthIndicator extends AbstractHealthIndicator`?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+
+  private final RestTemplate restTemplate;
+
+  @Autowired private Environment env;
```
CRoberto1926: You don't need to inject the `Environment` to get values for some properties:

```java
@Value("${camundaOutboxService.address}")
private String address;
@Value("${camundaOutboxService.port}")
private String port;
```

and so on for the other 4 properties

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+        .withDetail("Outbox Health", outboxHealth.build())
+        .withDetail("Kadai Health", kadaiHealth.build())
+        .build();
+  }
+
+  private ResponseEntity<EngineInfoRepresentationModel[]> pingCamundaRest() throws Exception {
+    String address = env.getProperty("camundaOutboxService.address");
+    String port = env.getProperty("camundaOutboxService.port");
+
+    String url = address + ":" + port + "/example-context-root/engine-rest/engine";
```
CRoberto1926: Prefer `UriBuilder`

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+      throw new Exception("Error connecting to Camunda REST API: " + e.getMessage(), e);
+    }
+  }
+
+  private ResponseEntity<EventCountRepresentationModel> pingOutBoxRestGetNumberFailedEvents()
+      throws Exception {
+    String address = env.getProperty("camundaOutboxService.address");
+    String port = env.getProperty("camundaOutboxService.port");
+
+    String url = address + ":" + port + "/example-context-root/outbox-rest/events/count?retries=0";
```
CRoberto1926: Prefer `UriBuilder`

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+          "Error getting number of failed events from Outbox: " + e.getMessage(), e);
+    }
+  }
+
+  private ResponseEntity<VersionInfoRepresentationModel> pingKadaiGetCurrentSchemaVersion()
+      throws Exception {
+    String address = env.getProperty("kadaiService.address");
+    String port = env.getProperty("kadaiService.port");
+
+    String url = address + ":" + port + "/kadai/api/v1/version";
```
CRoberto1926: Prefer `UriBuilder`

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+              HttpMethod.GET,
+              auth,
+              ParameterizedTypeReference.forType(VersionInfoRepresentationModel.class));
+      return response;
+    } catch (Exception e) {
+      throw new Exception("Error getting current schema version from Kadai: " + e.getMessage(), e);
+    }
+  }
+
+  public static HttpHeaders generateHeadersForUser(String user) {
```
CRoberto1926: This should be a private method right?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+    HttpHeaders headers = new HttpHeaders();
+    headers.add("Authorization", encodeUserAndPasswordAsBasicAuth(user));
+    headers.add("Content-Type", "application/ hal+json");
+    return headers;
+  }
+
+  public static String encodeUserAndPasswordAsBasicAuth(String user) {
```
CRoberto1926: This should be a private method right?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  private ResponseEntity<VersionInfoRepresentationModel> pingKadaiGetCurrentSchemaVersion()
+      throws Exception {
+    String address = env.getProperty("kadaiService.address");
+    String port = env.getProperty("kadaiService.port");
+
+    String url = address + ":" + port + "/kadai/api/v1/version";
+
+    HttpEntity<?> auth = new HttpEntity<>(generateHeadersForUser("admin"));
```
CRoberto1926: How is the method `pingKadaiGetCurrentSchemaVersion` supposed to work, if there isn't an "admin" user, or if the "admin" user has changed password, as every "admin" user in the world should do?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    return Health.status("External Services Status")
+        .withDetail("Camunda Health", camundaHealth.build())
+        .withDetail("Outbox Health", outboxHealth.build())
+        .withDetail("Kadai Health", kadaiHealth.build())
+        .build();
+  }
+
+  private ResponseEntity<EngineInfoRepresentationModel[]> pingCamundaRest() throws Exception {
```
CRoberto1926: Can't we extract just one method for `pingCamundaRest`, `pingOutBoxRestGetNumberFailedEvents` and `pingKadaiGetCurrentSchemaVersion`?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+  public static String encodeUserAndPasswordAsBasicAuth(String user) {
+    String toEncode = user + ":" + user;
+    return "Basic " + Base64.getEncoder().encodeToString(toEncode.getBytes(StandardCharsets.UTF_8));
+  }
+
+  public static class EngineInfoRepresentationModel {
+
+    private String name;
+
+    // Getter and setter for 'name'
```
CRoberto1926: Do we really need this comment?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
```diff
@@ -0,0 +1,199 @@
...
+    public void setName(String name) {
+      this.name = name;
+    }
+  }
+
+  public static class EventCountRepresentationModel {
+
+    private int eventsCount;
+
+    // Getter and setter for 'name'
```
CRoberto1926: Do we really need this comment?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: But if we want to call the current Schema Version of Kadai, we have to have admin permission

kadai-adapter-camunda-spring-boot-test/pom.xml
jamesrdi: Yes you're right, I will just use TestRestTemplate

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/HealthCheckTest.java
jamesrdi: Yes we can

kadai-adapter/pom.xml
jamesrdi: Yes you're right

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: I think name of component should match the endpoint of the health check, in this case "external-services" for it to work

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: I followed the guide to making health check [here](https://www.baeldung.com/spring-boot-health-indicators)

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: Yes you're right

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: Ok I'll change it

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: Ok I'll change it

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: Ok I'll change it

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: I need it for the test class, but I guess I can just copy the method to the test class

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: I need it for the test class, but I guess I can just copy the method to the test class

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: What do you mean?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: No we don't

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/HealthCheck.java
jamesrdi: No we don't

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
```diff
@@ -55,6 +55,14 @@ abstract class AbsIntegrationTest {
 
   protected static KadaiEngine kadaiEngine;
 
+  private boolean isIsInitialised() {
+    return isInitialised;
+  }
+
+  void setIsInitialised(boolean isInitialised) {
+    AbsIntegrationTest.isInitialised = isInitialised;
+  }
```
CRoberto1926: Either you make `isInitialised` not static, or you don't need these

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
```diff
@@ -0,0 +1,166 @@
...
+    Instant validRunTime = Instant.now().minus(Duration.ofMinutes(5));
+    when(spySchedulerHealthCheck.health())
+        .thenReturn(Health.up().withDetail("Last Run", validRunTime).build());
+  }
+
+  @AfterEach
+  void tearDown() {
+    if (mockServer != null) {
+      mockServer.reset();
+      mockServer = null;
```
CRoberto1926: Is .reset() not enough?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+  @AfterAll
+  void triggerSetUp() {
+    super.setIsInitialised(false);
+  }
+
+  @Test
+  void should_ReturnUp_When_CamundaServiceIsHealthy() {
+    mockServer
+        .expect(ExpectedCount.manyTimes(), requestTo("http://localhost:10020/engine-rest/engine"))
```
CRoberto1926: Are these URLs not defined in some application.properties? If yes, can we use those?

If they are not, can we at least define a constant for each URL and use it through the test classes?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.springframework.web.client.RestTemplate;
+
+@SpringBootTest(
+    classes = KadaiAdapterTestApplication.class,
+    webEnvironment = WebEnvironment.DEFINED_PORT)
+@AutoConfigureWebTestClient
+@ExtendWith(JaasExtension.class)
+@ContextConfiguration
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
CRoberto1926, line 49: Do you need these 2 annotations on this test class?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+    Instant validRunTime = Instant.now().minus(Duration.ofMinutes(5));
+    when(spySchedulerHealthCheck.health())
+        .thenReturn(Health.up().withDetail("Last Run", validRunTime).build());
+  }
+
+  @AfterEach
+  void tearDown() {
+    if (mockServer != null) {
+      mockServer.reset();
+      mockServer = null;
```
CRoberto1926: Is .reset() not enough?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+  void triggerSetUp() {
+    super.setIsInitialised(false);
+  }
+
+  @Test
+  void should_ReturnUp_When_OutboxServiceIsHealthy() {
+    mockServer
+        .expect(
+            ExpectedCount.manyTimes(),
+            requestTo("http://localhost:10020/outbox-rest/events/count?retries=0"))
```
CRoberto1926: Are these URLs not defined in some application.properties? If yes, can we use those?

If they are not, can we at least define a constant for each URL and use it through the test classes?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+  @Test
+  void should_ReturnDown_When_ReturnOtherStatusCode() {
+    mockServer
+        .expect(
+            ExpectedCount.manyTimes(),
+            requestTo("http://localhost:10020/outbox-rest/events/count?retries=0"))
+        .andExpect(method(HttpMethod.GET))
+        .andRespond(
+            withStatus(HttpStatus.ACCEPTED)
```
CRoberto1926, line 104: You could use [418](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418)

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+            "http://localhost:10020/actuator/health/external-services", String.class);
+
+    assertThat(response.getStatusCode()).isEqualTo(HttpStatus.SERVICE_UNAVAILABLE);
+    assertThat(response.getBody())
+        .contains("\"Outbox Health\":{\"status\":\"DOWN\"")
+        .contains("\"Outbox Service Error\":\"Unexpected status: 202 ACCEPTED\"");
+  }
+
+  @Test
+  void should_ThrowException_When_ReturnOtherStatusCode() {
```
CRoberto1926, line 118: What's the difference between this test and `should_ReturnDown_When_ReturnOtherStatusCode`?

kadai-adapter/src/main/java/io/kadai/adapter/impl/LastSchedulerRun.java
```diff
@@ -0,0 +1,17 @@
...
+
+@Component
+public class LastSchedulerRun {
+  private Instant lastRunTime;
+
+  public Instant getLastRunTime() {
+    return lastRunTime;
+  }
+
+  void touch() {
```
CRoberto1926, line 20: This method should also be called in `KadaiTaskTerminator`. And maybe we should also write an Archunit Test to be sure that each `@Scheduled` method calls `touch()`. Or some other solution to make sure that we don't forget `.touch()` in case new scheduled methods will be implemented

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
```diff
@@ -0,0 +1,135 @@
...
+    SchedulerHealthCheck schedulerHealthCheck = new SchedulerHealthCheck(spy);
+    Health health = Health.down().withDetail("Last Run", spy.getLastRunTime()).build();
+
+    assertThat(schedulerHealthCheck.health()).isEqualTo(health);
+  }
+
+  @Test
+  void should_ReturnDown_When_SchedulerNeverRuns() {
+    LastSchedulerRun spy = Mockito.spy(new LastSchedulerRun());
+    when(spy.getLastRunTime()).thenReturn(null);
```
CRoberto1926: This is actually the returning value, when `LastSchedulerRun.touch()` has never been called. This happens when no `@Scheduled` method has run. This does not mean that the scheduler is not working. I would suggest to initialize `LastSchedulerRun.lastRunTime` when instantiating the component object.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
```diff
@@ -0,0 +1,135 @@
...
+    SchedulerHealthCheck schedulerHealthCheck = new SchedulerHealthCheck(spy);
+    Health health = Health.down().withDetail("Scheduler Error", "Scheduler has never run").build();
+
+    assertThat(schedulerHealthCheck.health()).isEqualTo(health);
+  }
+
+  @Test
+  void should_ReturnDown_When_ExceptionThrown() {
+    LastSchedulerRun spy = Mockito.spy(new LastSchedulerRun());
+    Mockito.doThrow(new RuntimeException("Simulated exception")).when(spy).getLastRunTime();
```
CRoberto1926: In which real life scenario could `getLastRunTime()` throw an exception?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+  @Test
+  void should_ReturnUp_When_CallingHealthEndpoint() {
+    Instant validRunTime = Instant.now().minus(Duration.ofMinutes(5));
+    when(spyHealthCheck.health())
+        .thenReturn(Health.up().withDetail("Last Run", validRunTime).build());
+
+    ResponseEntity<String> response =
+        testRestTemplate.getForEntity(
+            "http://localhost:10020/actuator/health/external-services", String.class);
```
CRoberto1926: Same comment as the previous 2 for URLs in properties

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/TestCompletedTaskRetrieval.java
```diff
@@ -93,6 +95,9 @@ void should_CompleteCamundaTask_When_CompleteKadaiTask() throws Exception {
           this.camundaProcessengineRequester.getTaskFromHistoryFromTaskId(camundaTaskId);
       assertThat(taskRetrievalFromHistorySuccessful).isTrue();
     }
+    Instant lastRunTime = lastSchedulerRun.getLastRunTime();
+    assertThat(lastRunTime).isNotNull();
+    assertThat(lastRunTime).isAfter(Instant.now().minusSeconds(5));
```
CRoberto1926, line 100: What do you exactly want to test with these 3 lines? If the scheduler has correctly set the property?

kadai-adapter/pom.xml
```diff
@@ -55,6 +55,14 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-tx</artifactId>
         </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-actuator</artifactId>
+      </dependency>
+      <dependency>
+        <groupId>jakarta.ws.rs</groupId>
+        <artifactId>jakarta.ws.rs-api</artifactId>
```
CRoberto1926: why do we need this dependency?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
```diff
@@ -0,0 +1,36 @@
...
+    this.lastSchedulerRun = lastSchedulerRun;
+  }
+
+  @Override
+  public Health health() {
+    try {
+      Instant lastRunTime = lastSchedulerRun.getLastRunTime();
+
+      if (lastRunTime == null) {
+        return Health.down().withDetail("Scheduler Error", "Scheduler has never run").build();
```
CRoberto1926: See one of my comments above. This is at startup not an erro

CRoberto1926 (changes requested):

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
jamesrdi: We have discussed it and we want to use getter and setter to set isInitialised

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
jamesrdi: I don't think so, I tried it and it gave following error:

Caused by: java.lang.IllegalStateException: Cannot add more expectations after actual requests are made

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
jamesrdi: In the application.properties, we can only find the port number and the address, we cannot find the complete path, I guess we can define a constant for each URL

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
jamesrdi: Yes, 
DirtiesContext is used to clean the application context after running the health check test, so that other Tests will not be disturbed by previously existing context.
TestInstance is used so that I can use the @AfterAll method which calls the setIsInitialized from the AbsIntegrationTest. Without this, I cannot call methods from super class

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: I don't think so, I tried it and it gave following error:

Caused by: java.lang.IllegalStateException: Cannot add more expectations after actual requests are made

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: In the application.properties, we can only find the port number and the address, we cannot find the complete path, I guess we can define a constant for each URL

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: I cannot, because if I use that, it will be treated as an error and will be catched

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: This one catches any error thrown, but the other one explicitly does not allow any other status codes other than 200, e.g 202

kadai-adapter/src/main/java/io/kadai/adapter/impl/LastSchedulerRun.java
jamesrdi: Yes you are right, we can write a test in the ArchitectureTest to make sure that all scheduled methods call .touch()

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/TestCompletedTaskRetrieval.java
jamesrdi: Yes, I want to check if the scheduler correctly set the lastRunTime property

kadai-adapter/pom.xml
jamesrdi: It is for the UriBuilder

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
jamesrdi: Yes you're right, now that I think about it, it is unlikely that getLastRunTime() will throw an exception ,since it simply returns value of a lastRunTime, I will delete the test

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
jamesrdi: Yes, I will delete it

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
CRoberto1926: But why? The variable is static...

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
CRoberto1926: Alright, could you define constants please?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
CRoberto1926: Could you try again? Because on my machine everything is fine if I just use .reset(). We can even remove the null check.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
CRoberto1926: See my other comment: "it works on my machine"

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
CRoberto1926: See my other comment: please as constant in the class

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
CRoberto1926: Could you document this please? Either in the test method name, or with a comment on the tests.

kadai-adapter/pom.xml
CRoberto1926: I meant the one from Spring

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
jamesrdi: Yes you're right, I dont need the getters and setters

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaHealthCheckTest.java
jamesrdi: I have tried it again, it gives the same error when I install the module

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
jamesrdi: Yes you're right, I will implement it like you suggested

kadai-adapter/pom.xml
jamesrdi: Ok I will use the URI from java.net

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: I made another class specific for the endpoints, since the endpoints are repeated called also in other health check tests

CRoberto1926 (approved):
Good Job!

PR #109, opened 2025-04-16
Closes #99: Dedicated health checks for all @Scheduled tasks

### LINKED ISSUE #99: Dedicated health checks for all `@Scheduled` tasks
With #26 we introduced health-checks.
The response of `/actuator/health/external-services` currently looks like:
```json
{
    "status": "UP",
    "components": {
        // ...
        "Scheduler Health": {
            "status": "UP",
            "details": {
                "Last Run": "2025-02-03T10:20:36.174928Z"
            }
        }
    }
}
```

We want to extend the scheduler health - forming a boolean algebra over `@Scheduled` tasks and evaluated via conjunction.

jamesrdi:
### Thanks for your PR! Please fill out the following list :)

---

- [ ] I put the ticket or multiple tickets in review
- [x] Commit message format → Closes #&lt;Issue Number&gt; - Your commit message.
- [ ] Sonarcloud link : \<add the link here>
- [ ] No documentation update needed
- [x] Link to PR with documentation update: [\<add the link here>](https://github.com/kadai-io/kadai-doc/pull/88)
- [ ] No Release Notes needed
- [ ] Release Notes :

```

```

### DIFF
--- a/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
+++ b/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
@@ -4,7 +4,6 @@ import static io.kadai.adapter.integration.HealthCheckEndpoints.CAMUNDA_ENGINE_E
 import static io.kadai.adapter.integration.HealthCheckEndpoints.HEALTH_ENDPOINT;
 import static io.kadai.adapter.integration.HealthCheckEndpoints.OUTBOX_EVENTS_COUNT_ENDPOINT;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.when;
 import static org.springframework.test.web.client.match.MockRestRequestMatchers.method;
 import static org.springframework.test.web.client.match.MockRestRequestMatchers.requestTo;
 import static org.springframework.test.web.client.response.MockRestResponseCreators.withStatus;
@@ -15,8 +14,6 @@ import io.kadai.adapter.monitoring.SchedulerHealthCheck;
 import io.kadai.adapter.test.KadaiAdapterTestApplication;
 import io.kadai.common.test.security.JaasExtension;
 import io.kadai.common.test.security.WithAccessId;
-import java.time.Duration;
-import java.time.Instant;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -24,7 +21,6 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
@@ -60,12 +56,8 @@ class OutboxHealthCheckTest extends AbsIntegrationTest {
 
   @BeforeEach
   @WithAccessId(user = "admin")
-  void setUp() throws Exception {
+  void setUp() {
     mockServer = MockRestServiceServer.bindTo(restTemplate).ignoreExpectOrder(true).build();
-    SchedulerHealthCheck spySchedulerHealthCheck = schedulerHealthIndicator;
-    Instant validRunTime = Instant.now().minus(Duration.ofMinutes(5));
-    when(spySchedulerHealthCheck.health())
-        .thenReturn(Health.up().withDetail("Last Run", validRunTime).build());
   }
 
   @AfterEach
--- a/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
+++ b/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
@@ -1,21 +1,28 @@
 package io.kadai.adapter.integration;
 
 import static io.kadai.adapter.integration.HealthCheckEndpoints.HEALTH_ENDPOINT;
+import static io.kadai.adapter.integration.SchedulerHealthCheckTest.TestConfig.DUMMY_RUN_TIME;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.when;
 
 import io.kadai.adapter.impl.LastSchedulerRun;
 import io.kadai.adapter.monitoring.SchedulerHealthCheck;
+import io.kadai.adapter.monitoring.SchedulerHealthIndicator;
 import io.kadai.adapter.test.KadaiAdapterTestApplication;
 import io.kadai.common.test.security.JaasExtension;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mockito;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
@@ -36,16 +43,36 @@ class SchedulerHealthCheckTest extends AbsIntegrationTest {
 
   @TestConfiguration
   static class TestConfig {
+    public static final Instant DUMMY_RUN_TIME = Instant.now().minus(Duration.ofMinutes(5));
+
     @Bean
-    SchedulerHealthCheck schedulerHealthCheck() throws Exception {
-      LastSchedulerRun lastSchedulerRunMock = Mockito.mock(LastSchedulerRun.class);
-      Instant dummyRunTime = Instant.now().minus(Duration.ofMinutes(10));
-      when(lastSchedulerRunMock.getLastRunTime()).thenReturn(dummyRunTime);
+    SchedulerHealthCheck schedulerHealthCheck() {
+      LastSchedulerRun starterLastRun = mock(LastSchedulerRun.class);
+      LastSchedulerRun terminatorLastRun = mock(LastSchedulerRun.class);
+      LastSchedulerRun claimerLastRun = mock(LastSchedulerRun.class);
+      LastSchedulerRun completerLastRun = mock(LastSchedulerRun.class);
+      LastSchedulerRun cancelerLastRun = mock(LastSchedulerRun.class);
 
-      SchedulerHealthCheck realHealthCheck = new SchedulerHealthCheck(lastSchedulerRunMock);
-      SchedulerHealthCheck spyHealthCheck = Mockito.spy(realHealthCheck);
+      when(starterLastRun.getLastRunTime()).thenReturn(DUMMY_RUN_TIME);
+      when(terminatorLastRun.getLastRunTime()).thenReturn(DUMMY_RUN_TIME);
+      when(claimerLastRun.getLastRunTime()).thenReturn(DUMMY_RUN_TIME);
+      when(completerLastRun.getLastRunTime()).thenReturn(DUMMY_RUN_TIME);
+      when(cancelerLastRun.getLastRunTime()).thenReturn(DUMMY_RUN_TIME);
 
-      return spyHealthCheck;
+      HealthIndicator starter =
+          spy(new SchedulerHealthIndicator(starterLastRun, Duration.ofMinutes(10)));
+      HealthIndicator terminator =
+          spy(new SchedulerHealthIndicator(terminatorLastRun, Duration.ofMinutes(10)));
+      HealthIndicator claimer =
+          spy(new SchedulerHealthIndicator(claimerLastRun, Duration.ofMinutes(10)));
+      HealthIndicator completer =
+          spy(new SchedulerHealthIndicator(completerLastRun, Duration.ofMinutes(10)));
+      HealthIndicator canceler =
+          spy(new SchedulerHealthIndicator(cancelerLastRun, Duration.ofMinutes(10)));
+
+      SchedulerHealthCheck composite =
+          new SchedulerHealthCheck(starter, terminator, canceler, claimer, completer);
+      return spy(composite);
     }
   }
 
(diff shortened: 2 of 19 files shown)


kadai-adapter/src/main/java/io/kadai/adapter/monitoring/schedulers/KadaiTaskTerminatorSchedulerHealthCheck.java
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  @Override
+  public Health health() {
+    Instant lastRun = lastSchedulerRun.getLastRunTime();
+    return lastRun.isBefore(Instant.now().minus(Duration.ofMinutes(10)))
+        ? Health.down().withDetail("Last Run", lastRun).build()
+        : Health.up().withDetail("Last Run", lastRun).build();
+  }
+}
```
bruderj15: This is always the same.
Ideally use `@Bean` method definitions in a `@Configuration`class if possible.
If not, at least consider using a common super class overwriding `Health health()`.

bruderj15: I just saw `SchedulerLastRunConfig`, right direction.
Why all the redundancy with the explicit classes here then?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
```diff
@@ -48,6 +57,39 @@
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class OutboxHealthCheckTest extends AbsIntegrationTest {
+  @TestConfiguration
+  static class TestConfig {
```
CRoberto1926: Could you extract `TestConfig` in a separated files, and use it everywhere to avoid duplication of code? You are doing the same in each test file.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
```diff
@@ -60,55 +92,83 @@ void setUp() {
...
+  Stream<DynamicTest> should_ReturnUp_When_SchedulerRuns() throws Exception {
+    Stream<Pair<String, Class<? extends HealthIndicator>>> input =
+        Stream.of(
+            Pair.of("Kadai Task Starter", KadaiTaskStarterSchedulerHealthCheck.class),
+            Pair.of("Kadai Task Terminator", KadaiTaskTerminatorSchedulerHealthCheck.class),
+            Pair.of("Referenced Task Claimer", ReferencedTaskClaimerSchedulerHealthCheck.class),
+            Pair.of("Referenced Task Completer", ReferencedTaskCompleterSchedulerHealthCheck.class),
+            Pair.of(
+                "Referenced Task Claim Canceler",
+                ReferencedTaskClaimCancelerSchedulerHealthCheck.class));
```
CRoberto1926: This is the same also in the next `@TestFactory` right? You could extract it.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
```diff
@@ -60,55 +92,83 @@ void setUp() {
...
-    LastSchedulerRun spy = Mockito.spy(new LastSchedulerRun());
-    Instant invalidRunTime = Instant.now().minus(Duration.ofMinutes(15));
-    when(spy.getLastRunTime()).thenReturn(invalidRunTime);
-
-    SchedulerHealthCheck schedulerHealthCheck = new SchedulerHealthCheck(spy);
-    Health health = Health.down().withDetail("Last Run", spy.getLastRunTime()).build();
-
-    assertThat(schedulerHealthCheck.health()).isEqualTo(health);
+  @TestFactory
+  Stream<DynamicTest> should_ReturnDown_When_SchedulerRuns() throws Exception {
```
CRoberto1926: I think you could refactor this test and the previous one to use `@ParameterizedTest` with `@MethodSource`. In my opinion the tests would be easier to read. But I won't force this, you decide :)

kadai-adapter/src/main/java/io/kadai/adapter/impl/LastSchedulerRun.java
```diff
@@ -17,7 +15,7 @@ public Instant getLastRunTime() {
     return lastRunTime;
   }
 
-  void touch() {
+  public void touch() {
```
CRoberto1926: Why public?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/schedulers/KadaiTaskTerminatorSchedulerHealthCheck.java
CRoberto1926: I agree.

jamesrdi: Yes  you're right, I will change it

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: Yes, I will do it

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
jamesrdi: Yes

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/SchedulerHealthCheckTest.java
jamesrdi: Yes  I think Parameterized Test is better

kadai-adapter/src/main/java/io/kadai/adapter/impl/LastSchedulerRun.java
jamesrdi: You're right, it should not be public

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/OutboxHealthCheckTest.java
jamesrdi: Unfortunately I cannot, since if I separate the test config to separate file and try to use it by adding it to the list of classes in @ SpringBootTest or @ ContextConfiguration annotation, it will cause the test class to reload again the application context when running mvn install, which leads to "port 10020 is already used" problem

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
```diff
@@ -1,27 +1,52 @@
...
+          HealthIndicator referencedTaskCompleterSchedulerHealthCheck) {
+
+    schedulerContributors.put("Kadai Task Starter", kadaiTaskStarterSchedulerHealthCheck);
+    schedulerContributors.put("Kadai Task Terminator", kadaiTaskTerminatorSchedulerHealthCheck);
+    schedulerContributors.put(
+        "Referenced Task Claim Canceler", referencedTaskClaimCancelerSchedulerHealthCheck);
+    schedulerContributors.put("Referenced Task Claimer", referencedTaskClaimerSchedulerHealthCheck);
+    schedulerContributors.put(
+        "Referenced Task Completer", referencedTaskCompleterSchedulerHealthCheck);
   }
```
bruderj15, line 39: Consider having the beans injected here as it would greatly improve maintainability.
See [this](https://stackoverflow.com/questions/63658988/can-i-inject-map-of-all-impl-of-a-bean-in-spring) for inspiration.

kadai-adapter/src/main/java/io/kadai/adapter/configuration/AdapterConfiguration.java
```diff
@@ -43,35 +44,40 @@ public class AdapterConfiguration {
   @Bean
   @Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
```
bruderj15: `@Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)` is the default anyways, right?
Might as well just omit it here then.

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthIndicator.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public SchedulerHealthIndicator(LastSchedulerRun lastSchedulerRun, String name) {
+    this.lastSchedulerRun = lastSchedulerRun;
+    this.name = name;
+  }
+
+  @Override
+  public Health health() {
+    Instant lastRun = lastSchedulerRun.getLastRunTime();
+    return lastRun.isBefore(Instant.now().minus(Duration.ofMinutes(10)))
```
bruderj15: I think there's a prop for configuring how often scheduled jobs run, right?
Then the delta of 10min seems arbitrary here.
It should depend on the configured value.

kadai-adapter/src/main/java/io/kadai/adapter/configuration/AdapterConfiguration.java
CRoberto1926: I agree

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
```diff
@@ -1,27 +1,52 @@
...
+import org.springframework.boot.actuate.health.CompositeHealthContributor;
+import org.springframework.boot.actuate.health.HealthContributor;
 import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.boot.actuate.health.NamedContributor;
 import org.springframework.stereotype.Component;
 
-@Component
-public class SchedulerHealthCheck implements HealthIndicator {
-  private final LastSchedulerRun lastSchedulerRun;
+@Component("schedulerHealthCheck")
```
CRoberto1926: Why do you need to explicitly name it now?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
jamesrdi: Do you mean that I should just inject the dependencies as a map rather than individually injecting each HealthIndicator using `@Qualifier`?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthIndicator.java
jamesrdi: Yes you're right, I will change it

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
bruderj15: Yeah Map-injection is the way to go here.
Requires an additional type though.
I will take care of it with #121, therefore you can ignore it here.

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthCheck.java
bruderj15: Good question. I don't know why either.
But we'll need it for #121.

jamesrdi: You're right, I will just not use any name for this PR, and we can put the name later in the PR for #121

kadai-adapter/src/main/java/io/kadai/adapter/configuration/SchedulerHealthIndicatorConfig.java
```diff
@@ -0,0 +1,42 @@
...
+import io.kadai.adapter.monitoring.SchedulerHealthIndicator;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class SchedulerHealthIndicatorConfig {
+
+  @Bean("kadaiTaskStarterSchedulerHealthCheck")
```
bruderj15: Any reason you explicitly supply a bean-name?
Defaults to method name anyways afaik.
Might as well remove it then.

kadai-adapter/src/main/java/io/kadai/adapter/impl/LastSchedulerRun.java
```diff
@@ -2,9 +2,7 @@
 
 import jakarta.annotation.PostConstruct;
 import java.time.Instant;
-import org.springframework.stereotype.Component;
 
-@Component
 public class LastSchedulerRun {
```
bruderj15, line 7: this would greatly benefit from JavaDoc.
Something like: `A bean template for ...`

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthIndicator.java
```diff
@@ -0,0 +1,73 @@
...
+                      "kadai.adapter.scheduler.run.interval.for.complete.referenced.tasks.in."
+                          + "milliseconds",
+                      "10000")));
+    }
+    assert runInterval != null;
+    return lastRun.isBefore(Instant.now().minus(runInterval))
+        ? Health.down().withDetail("Last Run", lastRun).build()
+        : Health.up().withDetail("Last Run", lastRun).build();
+  }
+}
```
bruderj15: I cannot express how much I hate this.
You can come up with something better!

If you need help, hit me up! :)

bruderj15 (changes requested):

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/SchedulerHealthIndicator.java
bruderj15: I thought it was obvious, my bad.
Let me go into detail:
- This is a bean template, not yet a bean. Injecting the `Environment` is considered an anti-pattern because it - although the class being a bean template - may involve non-bean usage
- It being a bean template, why would it itself need to know about the required props for all the different indicators (implementations)?
- The check on name (a **custom!** runtime value) evades all compile- and bootup-time checks - making it unsafe
- It's also hard to maintain: Imagine adding twenty more actual beans (implementations) - the template would explode.

Without thinking about it in depth, the best solution probably is:
- to inject the environment value/default directly on-site at bean-construction (usage of constructor in `@Bean`-annotated method in a `@Configuration`)

kadai-adapter/src/main/java/io/kadai/adapter/configuration/SchedulerHealthIndicatorConfig.java
jamesrdi: It is used in the SchedulerHealthCheck class to inject each bean to the composite health contibutor

bruderj15: Yeah that doesn't change my previous statement, both:
```java
@Bean("kadaiTaskStarterSchedulerHealthCheck")
public HealthIndicator kadaiTaskStarterSchedulerHealthCheck(@Qualifier("kadaiTaskStarterLastRun") LastSchedulerRun lastRun) {
  return new SchedulerHealthIndicator(lastRun, "Kadai Task Starter");
}
```
and
```java
@Bean
public HealthIndicator kadaiTaskStarterSchedulerHealthCheck(@Qualifier("kadaiTaskStarterLastRun") LastSchedulerRun lastRun) {
  return new SchedulerHealthIndicator(lastRun, "Kadai Task Starter");
}
```
Will create a bean named `kadaiTaskStarterSchedulerHealthCheck`.

jamesrdi: I see now what you mean, yes you're right, I've deleted it

bruderj15:
#123 shipped major changes. Closing this in favor of a fresh one.

PR #136, opened 2025-06-11
Closes #124: Use Camunda System URLs in properties and allow multiple Camunda and Outbox for Health Check

### LINKED ISSUE #124: Use Camunda System URLs in properties and allow multiple Camunda and Outbox for Health Check
Currently the `CamundaHealthCheck` forces the url for the health-ping to be `.../engine-rest/engine`.

Make this more flexible by allowing to configure it - defaulting to the currently forced values.

jamesrdi:
## Definition of Done

    - will be squashed on merge
- [x] If the documentation needs an update, following there's a link to it: https://github.com/kadai-io/kadai-doc/pull/100
  - removed properties `camundaOutboxService.address`, `camundaOutboxService.port`, and `outbox.context-path` for getting URLs of Camunda and Outbox
  - No extra properties is needed now for configuring the Camunda and Outbox URLs for Health Check, the property `kadai-system-connector-camundaSystemURLs` is used instead to get the URLs, with the format `kadai-system-connector-camundaSystemURLs=<camundaSystem1-RestURL> | <camundaSystem1-OutboxRestURL> , ..., <camundaSystemN-RestURL> | <camundaSystemN-OutboxRestURL>`
  - Health checks for multiple Camunda and Outbox URLs is now possible
  - Group the Camunda and Outbox URLs according to the Camunda System they belong to, so it follow this structure:
  ```
  camundaSystems:
    camundaSystem1:
      camunda:
      outbox:
    camundaSystem2:
      camunda:
      outbox:
  ```
  - Possibility to toggle the health check for the entire composite of Camunda Systems on and off using the property `management.health.external-services.camunda-system.enabled=false/true` (default true)
  - Possibility to toggle the health check for all camunda or all outbox systems on and off using the property `management.health.external-services.camunda-system.camunda.enabled=false/true` and `management.health.external-services.camunda-systems.outbox.enabled=false/true` (default true)


bruderj15 (commented):
Let's think about the design for a second.

You introduced a composite for camunda and a one for outbox.
This is the distributive form (think ring).

What about factoring instead (undoing distribution)?

Then for `n` Camunda-Engines there would be `n` composites, each having two contributors:
- Outbox-Rest
- Camunda

I think this might represent the health of each system better, wouldn't it?

What do you think?

jamesrdi:
> Let's think about the design for a second.
> 
> You introduced a composite for camunda and a one for outbox. This is the distributive form (think ring).
> 
> What about factoring instead (undoing distribution)?
> 
> Then for `n` Camunda-Engines there would be `n` composites, each having two contributors:
> 
>     * Outbox-Rest
> 
>     * Camunda
> 
> 
> I think this might represent the health of each system better, wouldn't it?
> 
> What do you think?

I have changed the structure to follow the one that you suggested :)

bruderj15 (commented):
Whenever you provide Health-Detail-Keys, make sure they're valid JSON-Keys.
```java
Health.down()
          .withDetail("camundaEngines", e.getMessage())
          .withDetail("Base URL", urlString)
          .build();
```
should be
```java
Health.down()
          .withDetail("camundaEngines", e.getMessage())
          .withDetail("baseUrl", urlString)
          .build();
```
 There are quite a few of them, didn't want to open as many threads.

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/CamundaHealthIndicator.java
```diff
@@ -11,22 +11,13 @@
...
-      String camundaOutboxAddress,
-      int camundaOutboxPort,
-      String contextPath) {
+  public CamundaHealthIndicator(RestTemplate restTemplate, String urlString) {
     this.restTemplate = restTemplate;
-    this.camundaOutboxAddress = camundaOutboxAddress;
-    this.camundaOutboxPort = camundaOutboxPort;
-    this.contextPath = contextPath;
-    init();
+    this.urlString = urlString;
```
bruderj15, line 21: Do we even need to keep the string?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/CamundaHealthIndicator.java
```diff
@@ -36,25 +27,20 @@ public Health health() {
       CamundaEngineInfoRepresentationModel[] engines = response.getBody();
 
       if (engines == null || engines.length == 0) {
-        return Health.down().withDetail("camundaEngineError", "No engines found").build();
+        return Health.down()
+            .withDetail("camundaEngineError", "No engines found")
+            .withDetail("Base URL", urlString)
```
bruderj15: if you instead use the `url` like a few lines above for the happy-path, we may get rid of `urlString`

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/CamundaSystemsHealthComposite.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Map;
+import java.util.StringTokenizer;
+import org.springframework.boot.actuate.health.CompositeHealthContributor;
+import org.springframework.boot.actuate.health.HealthContributor;
+import org.springframework.boot.actuate.health.NamedContributor;
+import org.springframework.web.client.RestTemplate;
+
+public class CamundaSystemsHealthComposite implements CompositeHealthContributor {
+
+  private final Map<String, HealthContributor> healthContributors = new HashMap<>();
```
bruderj15, line 16: I wonder if we can just use a list instead. 
Does actuator allow this?

(Avoids the awkward naming of `camundaSystem[n]`)

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/ExternalServicesHealthComposite.java
```diff
@@ -29,29 +29,20 @@ public class ExternalServicesHealthComposite implements CompositeHealthContribut
...
-      @Value("${camundaOutboxService.port:8090}") int camundaOutboxPort,
-      @Value("${outbox.context-path:}") String contextPath,
       ReferencedTaskCompleter referencedTaskCompleter,
       ReferencedTaskClaimer referencedTaskClaimer,
       ReferencedTaskClaimCanceler referencedTaskClaimCanceler,
       KadaiTaskStarter kadaiTaskStarter,
-      KadaiTaskTerminator kadaiTaskTerminator) {
-    if (properties.getCamunda().getEnabled()) {
+      KadaiTaskTerminator kadaiTaskTerminator,
+      @Value("${kadai-system-connector-camundaSystemURLs}") String camundaSystemUrls) {
```
bruderj15: SpringBoot supports lists, doesn't it?

I'd prefer a list/array over custom notation.

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/OutboxHealthIndicator.java
```diff
@@ -35,31 +33,25 @@ public Health health() {
       ResponseEntity<OutboxEventCountRepresentationModel> response = pingOutBoxRest();
 
       if (response.getStatusCode() == HttpStatus.OK) {
-        return Health.up().withDetail("Outbox Service", response.getBody()).build();
+        return Health.up()
+            .withDetail("Outbox Service", response.getBody())
```
bruderj15: here are some more non-json keys

kadai-adapter/src/main/java/io/kadai/adapter/systemconnector/api/SystemConnector.java
```diff
@@ -93,6 +93,13 @@ public interface SystemConnector {
    */
   String getSystemUrl();
 
+  /**
+   * Get the URL of the outbox this connector connects to.
+   *
+   * @return the URL of the connected outbox.
+   */
+  String getSystemTaskEventUrl();
```
bruderj15: JavaDoc and Method-Name diverge

bruderj15:
Please do not force-push, makes iterative reviews tiresome.

Instead just append new commits and squash on merge.

jamesrdi:
> Please do not force-push, makes iterative reviews tiresome.
> 
> Instead just append new commits and squash on merge.

Ok, will do that from now on

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/CamundaHealthIndicator.java
jamesrdi: I want to just write the base path without the specific endpoint, so just localhost:8080/engine-rest or localhost:8080/outbox-rest, without the endpoints like /engine or /events

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/CamundaSystemsHealthComposite.java
jamesrdi: I dont think so, Spring Boot Actuator requires that each contributor must have a unique, human readable name.

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/ExternalServicesHealthComposite.java
jamesrdi: What do you mean by this? Are you suggesting that we change the structure of the property `kadai-system-connector-camundaSystemURLs`,if so, then it would contradict our aim which is to use the existing `kadai-system-connector-camundaSystemURLs` property as it is for the URLs in the health check

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/OutboxHealthIndicator.java
jamesrdi: Yes I will change them

kadai-adapter/src/main/java/io/kadai/adapter/systemconnector/api/SystemConnector.java
jamesrdi: I am going to delete this method becauase we dont need it anyways

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/ExternalServicesHealthComposite.java
bruderj15: No matter where we use the prop, it should really be a list, shouldn't it?

jamesrdi: I see what you mean, I have changed it

kadai-adapter-camunda-spring-boot-test/src/main/resources/application.properties
```diff
@@ -32,7 +32,7 @@ kadai.adapter.scheduler.run.interval.for.retries.and.blocking.taskevents.in.mill
 #
 # Set URLs of Camunda REST API and associated Camunda task events. The format is
 # <camundaSystem1-RestURL> | <camundaSystem1-TaskEventsURL> , ..., <camundaSystemN-RestURL> | <camundaSystemN-TaskEventsURL>
-kadai-system-connector-camundaSystemURLs=http://localhost:${server.port}/engine-rest/engine/default | http://localhost:${server.port}/outbox-rest
+kadai-system-connector-camundaSystemURLs=http://localhost:${server.port}/engine-rest | http://localhost:${server.port}/outbox-rest
```
bruderj15, line 35: Can we shine some light on this?
I am utterly confused.

`camundaSystemUrl` seems like an unfortunate name for a list right?
Here a `camundaSystemUrl` (element in the list `camundaSystemUrls`) itself is a `List<String>` isn't it?
Therefore `camundaSystemUrls` in fact is a `List<List<String>>` where the inner list always has size 2?

Seems a nightmare to process (for us: `StringTokenizer`) and to configure for clients.
Configuration would look like this?
```properties
kadai-system-connector-camundaSystemURLs={http://localhost:${server.port}/engine-rest1 | http://localhost:${server.port}/outbox-rest1, http://localhost:${server.port}/engine-rest2 | http://localhost:${server.port}/outbox-rest2}
```

This is awful, isn't it?
I cannot grasp all the URL-Config things related to camunda and outbox.
But: Is there any way such that supplying **one** URL per system suffices?
Do both urls per system share the identical prefix and only differ in their suffixes `camunda-rest` and `outbox-rest` (like in the example here)?

If they don't - so we really need to supply both urls per system (for multiple systems) - can we beautify the configuration with some spring-config mechanics?
Maybe with `@ConfigurationProperties`?

bruderj15 (commented):
This looks really good. Curious about the config stuff.

kadai-adapter-camunda-spring-boot-test/src/main/resources/application.properties
jamesrdi: Unfortunately, it is possible that the camunda and outbox url in a system do not share the same prefix, so they may use different server and different ports, but maybe I can try beautifying the configuration, I will look into it.

jamesrdi: But I do remember that we discussed we want to just use the existing `kadai-system-connector-camundaSystemURLs` property when configuring the URLs for Camunda and Outbox, so I dont think we need to beautify it now.

bruderj15: outsourced to #149

bruderj15, line 35: Is the removal of the suffix `/engine/default` breaking for our clients?

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/ExternalServicesHealthComposite.java
```diff
@@ -29,29 +30,20 @@ public class ExternalServicesHealthComposite implements CompositeHealthContribut
   public ExternalServicesHealthComposite(
       ExternalServicesHealthConfigurationProperties properties,
       RestTemplate restTemplate,
-      @Value("${camundaOutboxService.address:http://localhost}") String camundaOutboxAddress,
-      @Value("${camundaOutboxService.port:8090}") int camundaOutboxPort,
-      @Value("${outbox.context-path:}") String contextPath,
```
bruderj15, line 34: I think the changelog should clearly state:
- what properties have been removed
- how their targeted behavior is configured now

kadai-adapter-camunda-spring-boot-test/src/main/resources/application.properties
jamesrdi: I corrected it because /engine/default is an endpoint, not a base URL,  the property in kadai-adapter-camunda-spring-boot-example is correct, but not here

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/ExternalServicesHealthComposite.java
jamesrdi: Ok I will add it in the release notes

bruderj15: Alright, hit me up when it's done and I'll give the approval! :)

bruderj15 (approved):

kadai-adapter/src/main/java/io/kadai/adapter/monitoring/ExternalServicesHealthComposite.java
jamesrdi: It's done :)

PR #156, opened 2025-08-01
Closes #106: Migrate RestTemplate to RestClient

### LINKED ISSUE #106: Migrate `RestTemplate` to `RestClient`
This issue is the equivalent to [kadai-io#643](https://github.com/kadai-io/kadai/issues/643).

The project is currently using `RestTemplate`, which is deprecated in Spring Boot 3. The recommended alternative is `RestClient` (introduced in Spring 6) for synchronous requests or WebClient for asynchronous requests.

jamesrdi:
## Definition of Done

    - will be squashed on merge

### DIFF
--- a/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
+++ b/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
@@ -43,10 +43,9 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.boot.test.web.client.TestRestTemplate;
 import org.springframework.boot.test.web.server.LocalServerPort;
-import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.web.client.RestClient;
 
 /** Parent class for integrationtests for the KADAI-Adapter. */
 @ExtendWith(JaasExtension.class)
@@ -92,7 +91,7 @@ abstract class AbsIntegrationTest {
   @Resource(name = "camundaBpmDataSource")
   protected DataSource camundaBpmDataSource;
 
-  protected TestRestTemplate testRestTemplate;
+  protected RestClient restClient;
 
   @Autowired private ProcessEngineConfiguration processEngineConfiguration;
 
@@ -125,19 +124,17 @@ abstract class AbsIntegrationTest {
 
       isInitialised = true;
     }
-    this.testRestTemplate =
-        new TestRestTemplate(
-            new RestTemplateBuilder()
-                .rootUri("http://localhost:" + port)
-                .requestFactory(HttpComponentsClientHttpRequestFactory.class));
+    this.restClient = RestClient.builder()
+        .baseUrl("http://localhost:" + port)
+        .requestFactory(new HttpComponentsClientHttpRequestFactory())
+        .build();
     // set up camunda requester and kadaiEngine-Taskservice
     this.camundaProcessengineRequester =
         new CamundaProcessengineRequester(
             this.processEngineConfiguration.getProcessEngineName(),
-            this.testRestTemplate,
+            this.restClient,
             this.httpHeaderProvider);
-    this.kadaiOutboxRequester =
-        new KadaiOutboxRequester(this.testRestTemplate, this.httpHeaderProvider);
+    this.kadaiOutboxRequester = new KadaiOutboxRequester(this.restClient, this.httpHeaderProvider);
     this.taskService = kadaiEngine.getTaskService();
 
     // adjust polling interval, give adapter a little more time
--- a/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaProcessengineRequester.java
+++ b/kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaProcessengineRequester.java
@@ -24,11 +24,11 @@ import java.util.List;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
-import org.springframework.boot.test.web.client.TestRestTemplate;
-import org.springframework.http.HttpEntity;
-import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestClient;
 
 /** Class to assist with building requests against the Camunda REST API. */
 public class CamundaProcessengineRequester {
@@ -41,7 +41,7 @@ public class CamundaProcessengineRequester {
   private static final String PROCESS_INSTANCE_PATH = "/process-instance";
   private static final String TASK_PATH = "/task";
 
-  private final TestRestTemplate restTemplate;
+  private final RestClient restClient;
 
   private final String processEngineKey;
 
@@ -51,27 +51,25 @@ public class CamundaProcessengineRequester {
    * Constructor for setting up a requester for a process engine with a key other than "default".
    *
    * @param processEngineKey the key of the camunda process engine to be called.
-   * @param restTemplate the {@link TestRestTemplate} to be used for the REST calls.
+   * @param restClient the {@link RestClient} to be used for the REST calls.
    * @param httpHeaderProvider httpHeaderProvider to set correct headers on HTTP Request
    */
   public CamundaProcessengineRequester(
-      String processEngineKey,
-      TestRestTemplate restTemplate,
-      HttpHeaderProvider httpHeaderProvider) {
+      String processEngineKey, RestClient restClient, HttpHeaderProvider httpHeaderProvider) {
     this.processEngineKey = processEngineKey;
-    this.restTemplate = restTemplate;
+    this.restClient = restClient;
     this.httpHeaderProvider = httpHeaderProvider;
   }
 
   /**
    * Default constructor to use the default process engine with its key "default".
    *
-   * @param restTemplate the {@link TestRestTemplate} to be used for the REST calls.
+   * @param restClient the {@link RestClient} to be used for the REST calls.
    * @param httpHeaderProvider httpHeaderProvider to set correct headers on HTTP Request
    */
   public CamundaProcessengineRequester(
-      TestRestTemplate restTemplate, HttpHeaderProvider httpHeaderProvider) {
-    this("default", restTemplate, httpHeaderProvider);
+      RestClient restClient, HttpHeaderProvider httpHeaderProvider) {
+    this("default", restClient, httpHeaderProvider);
   }
 
   /**
@@ -91,11 +89,17 @@ public class CamundaProcessengineRequester {
             + PROCESS_DEFINITION_KEY_PATH
             + processKey
             + PROCESS_DEFINITION_START_PATH;
-    HttpEntity<String> requestEntity =
-        httpHeaderProvider.prepareNewEntityForCamundaRestApi("{" + variables + "}");
+    HttpHeaders headers = httpHeaderProvider.getHttpHeadersForCamundaRestApi();
+    String body = "{" + variables + "}";
 
     ResponseEntity<String> answer =
-        restTemplate.exchange(url, HttpMethod.POST, requestEntity, String.class);
+        restClient
+            .post()
+            .uri(url)
+            .headers(httpHeaders -> httpHeaders.addAll(headers))
+            .body(body)
+            .retrieve()
+            .toEntity(String.class);
     JSONObject processJson = new JSONObject(answer.getBody());
     return (String) processJson.get("id");
   }
@@ -114,10 +118,15 @@ public class CamundaProcessengineRequester {
     List<String> returnList = new ArrayList<String>();
 
     String url = BASIC_ENGINE_PATH + this.processEngineKey + TASK_PATH;
-    HttpEntity<Void> requestEntity = httpHeaderProvider.prepareNewEntityForCamundaRestApi();
+    HttpHeaders headers = httpHeaderProvider.getHttpHeadersForCamundaRestApi();
 
     ResponseEntity<String> answer =
-        this.restTemplate.exchange(url, HttpMethod.GET, requestEntity, String.class);
+        restClient
+            .get()
+            .uri(url)
+            .headers(httpHeaders -> httpHeaders.addAll(headers))
+            .retrieve()
+            .toEntity(String.class);
     JSONArray tasklistJson = new JSONArray(answer.getBody());
     for (int i = 0; i < tasklistJson.length(); i++) {
       JSONObject taskJson = (JSONObject) tasklistJson.get(i);
@@ -145,9 +154,15 @@ public class CamundaProcessengineRequester {
             + "/"
             + camundaTaskId
             + COMPLETE_TASK_PATH;
-    HttpEntity<String> requestEntity = httpHeaderProvider.prepareNewEntityForCamundaRestApi("{}");
+    HttpHeaders headers = httpHeaderProvider.getHttpHeadersForCamundaRestApi();
     ResponseEntity<String> answer =
-        this.restTemplate.exchange(url, HttpMethod.POST, requestEntity, String.class);
+        restClient
+            .post()
+            .uri(url)
+            .headers(httpHeaders -> httpHeaders.addAll(headers))
+            .body("{}")
+            .retrieve()
+            .toEntity(String.class);
     if (answer.getStatusCode().equals(HttpStatus.NO_CONTENT)) {
       // successful requests give no answer with a body
       return true;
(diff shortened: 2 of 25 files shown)


kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
```diff
@@ -125,19 +123,14 @@ void setUp() throws Exception {
 
       isInitialised = true;
     }
-    this.testRestTemplate =
-        new TestRestTemplate(
-            new RestTemplateBuilder()
-                .rootUri("http://localhost:" + port)
-                .requestFactory(HttpComponentsClientHttpRequestFactory.class));
```
bruderj15, line 132: What Request-Factory is used now? Is it's spec equivalent?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaProcessengineRequester.java
```diff
@@ -248,8 +284,18 @@ public boolean deleteProcessInstanceWithId(String processInstanceId, boolean ski
...
+      answer =
+          restClient
+              .delete()
+              .uri(url)
+              .headers(httpHeaders -> httpHeaders.addAll(requestEntity.getHeaders()))
+              .retrieve()
+              .toEntity(String.class);
+    } catch (HttpClientErrorException.NotFound e) {
+      answer = ResponseEntity.status(HttpStatus.NOT_FOUND).body(e.getResponseBodyAsString());
+    }
```
bruderj15: This seems awkward.
Is there a cleaner way?

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/KadaiOutboxRequester.java
```diff
@@ -105,8 +115,12 @@ public boolean setRemainingRetries(int id, int newRetries) throws JSONException
...
         httpHeaderProvider.prepareNewEntityForOutboxRestApi(
             "{\"remainingRetries\":" + newRetries + "}");
-    ResponseEntity<String> answer =
-        this.restTemplate.exchange(url, HttpMethod.PATCH, requestEntity, String.class);
+    ResponseEntity<String> answer = restClient.patch()
+        .uri(url)
+        .headers(httpHeaders -> httpHeaders.addAll(requestEntity.getHeaders()))
+        .body(requestEntity.getBody())
+        .retrieve()
+        .toEntity(String.class);
```
bruderj15: It seems like very often we construct the `requestEntity` but only end up using its fields instead.
I'd recommend to either:
- use the `requestEntity` directly for the request (if this is possible - I don't think so)
- avoid constructing the `requestEntity` and inlining the the values from previous usage.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/TestDisabledTaskClaim.java
```diff
@@ -115,11 +115,11 @@ void should_NotClaimOrCancelClaimCamundaTask_When_CamundaClaimingDisabled() thro
 
   private void setClaimingEnabled(boolean claimingEnbaled) throws Exception {
     Field claimerClaimingEnabled =
-        camundaTaskClaimer.getClass().getDeclaredField("claimingEnabled");
+        camundaTaskClaimer.getClass().getSuperclass().getDeclaredField("claimingEnabled");
     claimerClaimingEnabled.setAccessible(true);
     claimerClaimingEnabled.setBoolean(camundaTaskClaimer, claimingEnbaled);
     Field claimCancelerClaimingEnabled =
-        camundaTaskClaimCanceler.getClass().getDeclaredField("claimingEnabled");
+        camundaTaskClaimCanceler.getClass().getSuperclass().getDeclaredField("claimingEnabled");
```
bruderj15: What's happening here?

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/AbstractCamundaTaskClaimHandler.java
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestClient;
+
+/**
+ * Abstract base class for Camunda task claim operations. Provides common functionality for claiming
+ * and canceling claims of tasks.
+ */
+public abstract class AbstractCamundaTaskClaimHandler {
```
bruderj15: I highly endorse abstraction.

The diff is already quite big and hard to review.
Can we move this to a separate issue?

bruderj15 (commented):
That must've taken some time!

I've stopped the review with the last comment as I expect some downstream changes.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/AbsIntegrationTest.java
jamesrdi: You're right, it is not equivalent. I will change all restClient build to use an equivalent request factory

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaProcessengineRequester.java
jamesrdi: Unfortunately I dont think so, I agree it is complicated, especially with the try catch block, but restClient treats 404 as an error an throws it, which is why we have to manually catch it

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/KadaiOutboxRequester.java
jamesrdi: Unfortunately the method `getHttpHeadersForCamundaRestApi()` and `getHttpHeadersForOutboxRestApi()` are not public in the HttpHeaderProvider.java class, and therefore I cannot just create a HttpHeaders object and inline the header

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/AbstractCamundaTaskClaimHandler.java
jamesrdi: I actually need to make the abstraction here, else sonarqube quality gate will be red because of duplicate codes in CamundaTaskClaimer and CamundaTaskClaimCanceler

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/TestDisabledTaskClaim.java
jamesrdi: This is because I use the `AbstractCamundaTaskClaimHandler.java` to make abstraction, and now the CamundaTaskClaimer and CamundaTaskClaimCanceler extend the abstract class. So now the `claimingEnabled` attribute is not directly in the CamundaTaskClaimer and CamundaTaskClaimCanceler class, but they are in the super class `AbstractCamundaTaskClaimHandler.java`, which is why for the reflection to work, we get now the declared field from the superclass using `getSuperClass()`

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/AbstractCamundaTaskClaimHandler.java
bruderj15: The code has been duplicated for a while already, glad that you're tackling it!
Can we still move it into a separate Issue and PR though?

I dont mind the Quality-Gate failing for this PR here.

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/CamundaProcessengineRequester.java
bruderj15: What an awful design decision by Spring...

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/AbstractCamundaTaskClaimHandler.java
jamesrdi: Ok, I have removed it, and will solve it in a separate issue

kadai-adapter-camunda-spring-boot-test/src/test/java/io/kadai/adapter/integration/KadaiOutboxRequester.java
bruderj15: As we don't have any usage for the ` HttpEntity<String> requestEntity` anymore as the `RestClient` inlines its values, consider removing `prepareNewEntityForOutboxRestApi` and alike.
Then make mentioned getters public.

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/AbstractCamundaTaskClaimHandler.java
bruderj15: Moved to #159

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/CamundaTaskCompleter.java
```diff
@@ -56,9 +55,9 @@ public SystemResponse completeCamundaTask(
 
       return performCompletion(camundaSystemUrlInfo, referencedTask, requestUrlBuilder);
 
-    } catch (HttpStatusCodeException e) {
+    } catch (HttpClientErrorException e) {
```
bruderj15, line 59: We intentionally disregard 5xx now...?

kadai-adapter-camunda-system-connector/src/test/java/io/kadai/camunda/camundasystemconnector/acceptance/RetrieveCamundaTaskAccTest.java
```diff
@@ -48,17 +45,31 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  @AfterEach
+  void tearDown() {
+    try {
+      mockWebServer.shutdown();
+    } catch (Exception e) {
+      // ignore
```
bruderj15, line 65: Is this actually the recommended way?
No more need to supply the _now_ `RestClient`?

bruderj15: Ah an entire new dependency. There's no migration with the old one possible...?

kadai-adapter/src/test/java/io/kadai/adapter/monitoring/CamundaHealthIndicatorTest.java
```diff
@@ -16,56 +14,55 @@
...
-            ResponseEntity.ok()
-                .body(
-                    new CamundaEngineInfoRepresentationModel[] {
-                      new CamundaEngineInfoRepresentationModel()
-                    }));
+    CamundaEngineInfoRepresentationModel[] engines =
+        new CamundaEngineInfoRepresentationModel[] {new CamundaEngineInfoRepresentationModel()};
+    ResponseEntity<CamundaEngineInfoRepresentationModel[]> response = ResponseEntity.ok(engines);
+
+    Mockito.doReturn(response).when(camundaHealthIndicatorSpy).pingCamundaRest();
```
bruderj15: This is flattening the test. 
I'd still prefer it to mock camundas response, not our utility method.

kadai-adapter-camunda-system-connector/src/test/java/io/kadai/camunda/camundasystemconnector/acceptance/RetrieveCamundaTaskAccTest.java
reynaldj: Yes we dont need to supply RestClient anymore. Unlike MockRestServiceServer which intercepts RestTemplate calls, MockWebServer is a real HTTP server that responds to HTTP requests. When I pass the MockWebServer URL to taskRetriever, it makes HTTP calls to that URL and receives the mock responses enqueued. 

Sadly we cannot use MockRestServiceServer for RestClient. I will also delete the unused RestClient.

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/CamundaTaskCompleter.java
jamesrdi: You're right, I will also catch the 5xx error

kadai-adapter-camunda-system-connector/src/test/java/io/kadai/camunda/camundasystemconnector/acceptance/RetrieveCamundaTaskAccTest.java
jamesrdi: Yes we dont need to supply RestClient anymore. Unlike MockRestServiceServer which intercepts RestTemplate calls, MockWebServer is a real HTTP server that responds to HTTP requests. When I pass the MockWebServer URL to taskRetriever, it makes HTTP calls to that URL and receives the mock responses enqueued.
Sadly we cannot use MockRestServiceServer for RestClient. I will also delete the unused RestClient.

kadai-adapter/src/test/java/io/kadai/adapter/monitoring/CamundaHealthIndicatorTest.java
jamesrdi: You're right, I will change the test to mock camundas response, also for the outbox

kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/CamundaTaskClaimCanceler.java
```diff
@@ -75,26 +75,31 @@ public SystemResponse cancelClaimOfCamundaTask(
...
           return new SystemResponse(HttpStatus.OK, null);
         } else {
-          LOGGER.warn("Caught Exception when trying to cancel claim camunda task", e);
+          LOGGER.warn("Caught client error when trying to cancel claim camunda task", e);
           throw e;
         }
+      } catch (HttpServerErrorException e) {
+        // Handle 5xx errors (server errors - should be retried)
+        LOGGER.warn("Caught server error when trying to cancel claim camunda task, will retry", e);
+        throw e;
```
bruderj15: Catch-Log-Throw is an anti-pattern.
By previously catching `HttpStatusCodeException`, both 4xx and 5xx were mapped to a `SystemResponse`.
Therefore we would change behavior here.

NB: You're right that - if done right - 4xx and 5xx _should_ be handled differently. That's not the scope of this PR though. Consider opening an issue.

kadai-adapter-camunda-system-connector/src/test/java/io/kadai/camunda/camundasystemconnector/acceptance/RetrieveCamundaTaskAccTest.java
```diff
@@ -168,13 +171,12 @@ void should_GetFinishedCamundaTask() {
...
 
-    String camundaSystemUrl = "http://localhost:8080";
+    String camundaSystemUrl = mockWebServer.url("").toString();
     String camundaSystemEngineIdentifier = "default";
-    mockServer
-        .expect(requestTo(camundaSystemUrl + "/events?type=complete&type=delete"))
-        .andExpect(method(HttpMethod.GET))
-        .andExpect(content().contentType(org.springframework.http.MediaType.APPLICATION_JSON))
-        .andRespond(withSuccess(expectedReplyBody, MediaType.APPLICATION_JSON));
+    mockWebServer.enqueue(
```
bruderj15, line 166: I think using a queue makes the tests hard to understand.
Also `String camundaSystemUrl = mockWebServer.url("").toString()` seems off-putting.

Does the dependency instead provide a way to simulate the _mock-specific-target-with-response_ behavior?

bruderj15 (commented):
With the recent commit we changed the `RetrieveCamundaTaskAccTest` from an integration-test to a unit-test.
Its name says: Acceptance-Test.
Therefore it _should_ cover a whole path (entire slice) of the backend.

Any particular reason for the change?

jamesrdi:
> With the recent commit we changed the `RetrieveCamundaTaskAccTest` from an integration-test to a unit-test. Its name says: Acceptance-Test. Therefore it _should_ cover a whole path (entire slice) of the backend.
> 
> Any particular reason for the change?

You are right. Then I will just revert the change, according to [this](https://codefarm0.medium.com/testing-restclient-in-spring-boot-3-x-897a60fd2fd2), best practice is to use MockWebServer for integration test, so I will just do that.

bruderj15 (commented):
Woah, would've had the same remarks as Copilot.
Techonologia!

bruderj15 (commented):
CI failing :(

bruderj15 (approved):
For the record: I opened https://github.com/kadai-io/KadaiAdapter/issues/191

bruderj15 (approved):

PR #179, opened 2025-08-27
Camunda8 MVP

leantios:
The workflow of loading a newly created task from C8 to Kadai has been tested. The current solution with the systemURL is not very nice, but there is an improvement ticket: https://github.com/kadai-io/KadaiAdapter/issues/203

## Definition of Done

- [ ] The corresponding ticket is in state `In Review`
- [ ] The corresponding ticket is attached to this Pull-Request
    - will be squashed on merge


kadai-adapter-camunda-8-spring-boot-example/src/main/java/io/kadai/adapter/KadaiAdapterApplication.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+
+/** Application that provides an adapter between KADAI and one or more external systems. */
+@SpringBootApplication
+@EnableScheduling
+@EnableTransactionManagement
+public class KadaiAdapterApplication {
```
bruderj15: Can we infix this with `Camunda8`?

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/HttpComponentsClientProperties.java
```diff
@@ -0,0 +1,31 @@
+package io.kadai.adapter.systemconnector.camunda.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "httpcomponentsclient")
+public class HttpComponentsClientProperties {
```
bruderj15: Spring does not allow to configure this out of the box?

bruderj15 (commented):
I fear the `todo`s to silently get swallowed.
I'd recommend to explicitly reference the corresponding github issues right in the code at the `todo`s.

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/HttpComponentsClientProperties.java
leantios: ´ClientHttpRequestFactorySettings´ is deprecated

bruderj15: The official docs for latest spring-boot@v3.5.5 don't state so: https://docs.spring.io/spring-boot/api/java/org/springframework/boot/http/client/ClientHttpRequestFactorySettings.html

bruderj15: The _plain old_ constructors are because they moved to a `record-class` instead.
The class itself is still to be used.

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/Camunda8SystemUrls.java
```diff
@@ -0,0 +1,17 @@
+package io.kadai.adapter.systemconnector.camunda.config;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class Camunda8SystemUrls {
```
bruderj15: What's the intention here?
Is it required as placeholder for usage somewhere?

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -0,0 +1,140 @@
...
+      Object variableObj = job.getVariablesAsMap().get(variableName);
+      if (variableObj instanceof String variableAsString && !variableAsString.isBlank()) {
+        return variableAsString;
+      }
+      if (variableObj instanceof Integer variableAsInt) {
+        return String.valueOf(variableAsInt);
+      }
+    } catch (Exception e) {
+      LOGGER.warn(
+          "Caught exception while trying to retrieve {} for task {} in ProcessDefinition {}",
```
bruderj15:
```suggestion
          "Caught exception while trying to retrieve '{}' for task '{}' in ProcessDefinition '{}'",
```

Alike for the others :)

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -0,0 +1,140 @@
...
+        return "{}";
+      }
+
+      Map<String, Object> variables = new HashMap<>();
+
+      variableNames.forEach(
+          nameOfVariableToAdd ->
+              variables.put(nameOfVariableToAdd, getVariable(job, nameOfVariableToAdd)));
+
+      ObjectMapper objectMapper = new ObjectMapper();
```
bruderj15: Construction of `ObjectMapper` is quite expensive afaik.
Consider a class-bound or static instance.

NB: Careful, often in Spring-Boot one declares it as a bean which may be undesireable here. Should use a static one instead.

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -0,0 +1,140 @@
...
+  private String getVariable(ActivatedJob job, String variableName) {
+    try {
+      Object variableObj = job.getVariablesAsMap().get(variableName);
+      if (variableObj instanceof String variableAsString && !variableAsString.isBlank()) {
+        return variableAsString;
+      }
+      if (variableObj instanceof Integer variableAsInt) {
+        return String.valueOf(variableAsInt);
+      }
+    } catch (Exception e) {
```
bruderj15: Does it really need to be the top-level `Exception`?

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCompletion.java
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      if (!gotActivated) {
+        gotActivated = true;
+        if (LOGGER.isInfoEnabled()) {
+          LOGGER.info(
+              "UserTaskListener kadai-receive-task-completed-event activated successfully, "
+                  + "connected to process instance {}",
+              job.getProcessInstanceKey());
+        }
+      }
```
bruderj15: This looks weird. It's simply for informational purpose?
Consider something like `@PostConstruct` to separate business-logic (job) from tracing.

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
leantios: No it does not, good point

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/HttpComponentsClientProperties.java
leantios: You are correct, I was looking at `org.springframework.boot.web.client.ClientHttpRequestFactorySettings` before, which is deprecated. However, I think we cannot replicate the current solution with less code, as `ConfigurationProperties` reads the `connectionTimeout` and `readTimeout` from the application.properties and returns a _default_ value if they are not present. Do you have a way to replicate this behaviour in mind?

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/Camunda8SystemUrls.java
leantios: Yes, it is a placeholder. However I removed it due to YAGNI

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCompletion.java
leantios: I did it the same way as in `io.kadai.adapter.camunda.tasklistener.KadaiTaskListener`. I don't want to see the info message when the class is initialized, but when Camunda calls my worker for the first time. However, I can remove it, if you do not like it

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/HttpComponentsClientProperties.java
bruderj15: I was hoping it was auto-configurable, unfortunately it's not.
Your approach is right then.

But I'd like to avoid the duplication of the class as it already exists here:
https://github.com/kadai-io/KadaiAdapter/blob/5c9c6e02d4055bea71460ef394caae89ec879a7e/kadai-adapter-camunda-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/HttpComponentsClientProperties.java

Consider moving it to a shared package.

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCompletion.java
bruderj15: I'd either:
- always info-log the invocation (totally fine and recommended from me)
- never log it

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
bruderj15: You changed it for the others but it's still missing here, innit?

kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCompletionTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ContextConfiguration;
+
+@Disabled
+@ContextConfiguration(classes = {Camunda8ConnectorTestConfiguration.class})
+@SpringBootTest
+@CamundaSpringProcessTest
```
bruderj15: As discussed in https://github.com/kadai-io/KadaiAdapter/issues/178#issuecomment-3252960852, remove this.

kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/config/Camunda8ConnectorTestConfiguration.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.kadai.adapter.impl.service.KadaiTaskCompletionServiceImpl;
+import io.kadai.adapter.manager.AdapterManager;
+import io.kadai.adapter.systemconnector.camunda.tasklistener.UserTaskCompletion;
+import io.kadai.adapter.systemconnector.camunda.tasklistener.util.ReferencedTaskCreator;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class Camunda8ConnectorTestConfiguration {
```
bruderj15: Also remove this

kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCompletionTest.java
leantios: As mentioned in https://github.com/kadai-io/KadaiAdapter/issues/178#issuecomment-3253450972 I do indeed think that we need ProcessTests. However, I agree that a disabled test should not be part of the MVP.

kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/config/Camunda8SystemConnectorConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+        .readTimeout(Duration.ofMillis(httpComponentsClientProperties.getReadTimeout()))
+        .requestFactory(HttpComponentsClientHttpRequestFactory.class)
+        .build();
+  }
+
+  @Bean
+  Integer getFromKadaiToAdapterBatchSize(
+      @Value("${kadai.adapter.sync.kadai.batchSize:#{64}}") final Integer batchSize) {
+    return batchSize;
+  }
```
bruderj15, line 40: This is weird. Any component could just use `@Value`.
Or are there some config-boundaries?
If yes, then it screams for config-leakage - suggesting to rethink design.

leantios: `KadaiSystemConnectorImpl` fetches the batch size as a bean:
```
Integer batchSize = AdapterSpringContextProvider.getBean(Integer.class);
```
I don't know the idea behind that implementation. Should we open an extra ticket? I don't see how this is coupled to Camunda8. Maybe you could give more details on what exactly you mean with config leakage.

bruderj15: omg this is incredibly bad

bruderj15: Yeah separate ticket

bruderj15: Opened https://github.com/kadai-io/KadaiAdapter/issues/190

bruderj15 (approved):

Issue #192, opened 2025-09-09
Refactor logic-leaking Health-Check


bruderj15:
Currently very bad structure for health-check.
Package `io.kadai.adapter.monitoring` in module `kadai-adapter` contains connector-specific health-indicators:
- `CamundaHealthIndicator`
- `CamundaOutboxHealthComposite`
- `CamundaaSystemsHealthComposite`
- `OutboxHealthIndicator`

These are not core-related - not inherently part of the adapter but **only** implied by the C7-plugin!

Instead these misfits should be located in the module of the c7-plugin (system-connector impl).
The module `kadai-adapter` (kernel) must provide easy integration with system-connector (plugin) health-checks.

bruderj15:
Might as well also adjust the `kadai-adapter/ArchitectureTest`. It should certainly have covered this.

bruderj15:
This is inherently in scope with #118 due to #185.

leantios:
But we can implement this ticket independently of the ongoing C8 implementation, right? Then I would suggest removing the parent issue

bruderj15:
Although in general I'd agree I think it's fine here.
Changing the health-check will be a breaking change.
As C8 is the only feature in-sight - introducing breaking changes, probably released with version 11 - they're a good fit imo.

leantios:
For the breaking architecture changes (https://github.com/kadai-io/KadaiAdapter/issues/162, https://github.com/kadai-io/KadaiAdapter/issues/161), we also merged into the [releases/v11.0.0](https://github.com/kadai-io/KadaiAdapter/tree/releases/v11.0.0) branch and did not handle them as sub-issues to the C8 issue.

bruderj15:
Buying it then

PR #194, opened 2025-09-10
C8: Closes #149 - Spring-driven configuration of system-url for C7 & C8

### LINKED ISSUE #149: Spring-Driven configuration of SystemUrls for C7
When configuring camunda system-connectors, for each camunda system we need to supply the:
- Engine REST-Url
- Outbox REST-Url

It currently looks like this in `application-properties`:
```properties
kadai-system-connector-camundaSystemURLs=
  http://localhost:${server.port}/engine-rest1 | http://localhost:${server.port}/outbox-rest1,
  http://localhost:${server.port}/engine-rest2 | http://localhost:${server.port}/outbox-rest2
```

Here a camundaSystemUrl (element of the list `camundaSystemURLs`) itself is a `List<String>`.
Therefore `camundaSystemURLs` in fact is a `List<List<String>>` where the inner list always has size 2.
This is bad.

It not only is painful to read for humans, but also for machines as it requires us to parse the properties String ourselves (non-standard format: heavy anti-pattern).

Instead we should use a `List<CamundaSystemConfig>` with `@ConfigurationProperties` where `CamundaSystemConfig` encapsulates both required URLs.

EDIT: Curently this property also seems to be required to run Camunda8 - probably due to dependenc on it by the health-check for C7. Fix this as well.

bruderj15:
This has mandatory implications on #192. Linked in-code via `TODO`.

## Definition of Done

- [ ] The corresponding ticket is in state `In Review`
    - will be squashed on merge


bruderj15:
Diff is broken, GitHub-Bug...?

PR #225, opened 2025-09-29
Closes #224 - Use Constructor Injection

### LINKED ISSUE #224: C8: Use Constructor Injection
Currently, in `Camunda8SystemConnectorConfiguration`, we directly create an instance of `KadaiTaskStarterServiceImpl` without letting spring Autowire it:

```
  @Bean
  @Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
  UserTaskCreation userTaskListenerCreation(final AdapterManager adapterManager) {
    return new UserTaskCreation(
        new KadaiTaskStarterServiceImpl(adapterManager),
        new ReferencedTaskCreator());
  }
```
This leads to the condition that sometimes, the `runAsUser` (uses the `@Value` annotation) is not loaded into `KadaiTaskStarterServiceImpl`, depending on the order in which Spring initializes the classes. Without the `runAsUser`, we don't have access to the database.

Generally, constructor injection is the best practice here and also helps us fixes the bug described above.

leantios:
## Definition of Done

    - will be squashed on merge


bruderj15 (approved):
Well done!

PR #232, opened 2025-10-01
Closes #219 - C8: Add direct failure if an exception is thrown in a Job Worker 

### LINKED ISSUE #219: C8: Add direct failure if an exception is thrown in a Job Worker
Just rethrow the exception.

lucalimbach:
## Definition of Done

    - will be squashed on merge


bruderj15 (approved):
For historical purposes:
Errors are handled by the Job-Worker via Camunda/Spring.
Their handling simply logs the error and fails the job.
No need for us to do it too.

leantios (approved):

PR #307, opened 2025-11-19
Closes #253 - Resolve SoC-violation for dependency-versions being listed in root-parent

### LINKED ISSUE #253: `kadai-adapter-parent/pom.xml` violates SoC as it contains dep-versions for plugin-deps

bruderj15:
## Definition of Done

    - will be squashed on merge

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,22 @@ updates:
       timezone: Europe/Berlin
     open-pull-requests-limit: 99
 
+  - package-ecosystem: maven
+    directory: "/kadai-adapter-camunda7-parent"
+    schedule:
+      interval: daily
+      time: "06:00"
+      timezone: Europe/Berlin
+    open-pull-requests-limit: 99
+
+  - package-ecosystem: maven
+    directory: "/kadai-adapter-camunda8-parent"
+    schedule:
+      interval: daily
+      time: "06:00"
+      timezone: Europe/Berlin
+    open-pull-requests-limit: 99
+
   - package-ecosystem: github-actions
     directory: "/"
     schedule:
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -8,6 +8,22 @@ updates:
...
 
+  - package-ecosystem: maven
+    directory: "/kadai-adapter-camunda7-parent"
+    schedule:
+      interval: daily
+      time: "06:00"
+      timezone: Europe/Berlin
+    open-pull-requests-limit: 99
+
+  - package-ecosystem: maven
```
leantios, line 19: Are the sub-modules of the camunda8 package then also automatically managed by dependabot?

bruderj15: Automatically?
Dependabot inspects the `pom.xml` we specify in the root of given ecosystem.
For us these are
- Project-Root (contains most dep-versions)
- C7-Plugin-Parent (contains few C7 dep-versions)
- C8-Plugin-Parent (contains few C8 dep-versions)

Dependabot does not inspect nested modules.
That's why we now need to explicitly list the C7 and C8-parent.
As all dep-versions are listed in any of these, we're covering every dep.

kadai-adapter-camunda8-parent/pom.xml
```diff
@@ -16,11 +16,19 @@
...
   </properties>
 
   <modules>
     <module>kadai-adapter-camunda-8-system-connector</module>
     <module>kadai-adapter-camunda-8-spring-boot-example</module>
   </modules>
 
+  <repositories>
+    <repository>
+      <id>camunda-bpm-nexus</id>
```
leantios: I think this is only necessary for C7, no?

bruderj15: Right you are!

leantios (approved):

Issue #211, opened 2025-09-19
Resolve heavy SoC-Viloation in core-module `kadai-adapter` related to `ReferencedTask`


bruderj15:
## Context

The module `kadai-adapter` acts as the kernel (core) in the micro-kernel architecture of the KadaiAdapter.
Example extensions are the C7 and C8 connectors. These are _plug-ins_.

The core therefore shall obviously never know anything about any particular plugin at all - besides implementation for the plugin-contract (interfaces `InboundSystemConnector` & `OutboundSystemConnector`.

However, it currently does. 

## Problem

The `ReferencedTask` is a reference to an external task - one of a _plugged-in_ system, e.g. C7.
Currently the type `ReferencedTask` contains fields for a specific plugin - C7:

https://github.com/kadai-io/KadaiAdapter/blob/646638d21b8441c6cd171e44d4dd311607fc4b07/kadai-adapter/src/main/java/io/kadai/adapter/systemconnector/api/ReferencedTask.java#L24-L28

It also uses these fields in the orchestration-logic inside the core.
The C7-Implementation-Details (usage of the outbox) are therefore leaking into the core!
This is horrendeous.

## Goals and possible solution

- The core `ReferencedTask` only contains [core-relevant](https://github.com/kadai-io/KadaiAdapter/blob/master/kadai-adapter-kadai-connector/src/main/java/io/kadai/adapter/kadaiconnector/api/impl/TaskInformationMapper.java) fields
- If extensions require extra fields, these should be generalized - such as an ID of some external task
- Consider turning `ReferencedTask` into an interface

leantios:
Note: `ReferencedTask.taskState` is also relevant, it is used in `KadaiSystemConnectorImpl::terminateKadaiTask`. https://github.com/kadai-io/KadaiAdapter/issues/213 wants to improve clearness by using an Enum for the possible states

PR #312, opened 2025-11-24
Closes #239: Setting Due/Planned while creating Tasks in KADAI via Adapter

### LINKED ISSUE #239: Setting Due/Planned while creating Tasks in KADAI via Adapter
When synchronizing tasks from camunda to KADAI, adjust the logic of setting due/planned according to this table:

Image

jamesrdi:
## Definition of Done

    - will be squashed on merge


kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -24,6 +26,9 @@ public class ReferencedTaskCreator {
 
   private final Camunda8System camunda8System;
 
+  @Value("${kadai.servicelevel.validation.enforce:false}")
```
bruderj15, line 30: Instead inject `CamundaListenerConfiguration`

bruderj15 (commented):
I really like the testing!
Could be improved even further by making the tests parameterized.

We're duplicating the business-logic here. This is a big no-no.
Try to get it to a single point.

NB: The amount and wording of comments makes it seem this is entirely vibe-coded. Might want to get rid off this.

jamesrdi:
> I really like the testing! Could be improved even further by making the tests parameterized.
> 
> We're duplicating the business-logic here. This is a big no-no. Try to get it to a single point.
> 
> NB: The amount and wording of comments makes it seem this is entirely vibe-coded. Might want to get rid off this.

What do you mean getting it to a single point?

bruderj15:
> What do you mean getting it to a single point?

The logic for due/planned is currently scattered. There are multiple places where `if due null && ...` is repeated.
I recommend moving this into a single place and then just reuse it.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
jamesrdi: But the module kadai-adapter-8-camunda-system-connector does not depend on the module kadai-adapter-camunda-listener in kadai-adapter-camunda7-parent, which contains the CamundaListenerConfiguration

kadai-adapter/src/main/java/io/kadai/adapter/util/DateTimeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private DateTimeUtils() {}
+
+  public static String formatDate(Date date) {
+    return date == null ? null : FORMATTER.format(date.toInstant());
+  }
+
+  public static String formatOffsetDateTime(OffsetDateTime odt) {
+    return odt == null ? null : odt.format(FORMATTER);
+  }
```
bruderj15: this can probably be generalized over some interface

kadai-adapter/src/main/java/io/kadai/adapter/util/DateTimeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * Shared helper class for date/time formatting and calculation of planned/due.
+ */
+public final class DateTimeUtils {
```
bruderj15: Separation of concerns badly violated here.
The util-class:
- formats time-data
- computes time-data based on domain-knowledge

Big nono. The domain part does not belong here!

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -35,15 +40,36 @@ public ReferencedTaskCreator(Camunda8System camunda8System) {
...
+                + ". "
+                + "This is not allowed when kadai.servicelevel.validation.enforce is true.");
+      } else {
+        referencedTask.setPlanned(pd.planned);
+        referencedTask.setDue(pd.due);
+      }
+    } else {
+      referencedTask.setPlanned(pd.planned);
+      referencedTask.setDue(pd.due);
+    }
```
bruderj15: this can be flattened

kadai-adapter/src/main/java/io/kadai/adapter/util/DateTimeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+      String planned = formatter.apply(followUp);
+      return new PlannedDue(planned, null, false);
+    } else if (dueSet) {
+      String dueStr = formatter.apply(due);
+      return new PlannedDue(null, dueStr, false);
+    } else {
+      String planned = formatter.apply(nowSupplier.get());
+      return new PlannedDue(planned, null, false);
+    }
+  }
```
bruderj15: This is a great example on how separation of logic and representation can greatly reduce complexity.
See, this code is very readable - compare it to the previous versions!
However, you can do even better by moving the formatting part out of the logic.

kadai-adapter/src/main/java/io/kadai/adapter/util/DateTimeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public static String formatOffsetDateTime(OffsetDateTime odt) {
+    return odt == null ? null : odt.format(FORMATTER);
+  }
+
+  private static <T> PlannedDue computePlannedDueGeneric(
+      T followUp,
+      T due,
+      Function<T, String> formatter,
+      Supplier<T> nowSupplier
```
bruderj15: There's no common interface that abstracts over `T.now()`?

kadai-adapter/src/main/java/io/kadai/adapter/util/DateTimeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public static String formatOffsetDateTime(OffsetDateTime odt) {
+    return odt == null ? null : odt.format(FORMATTER);
+  }
+
+  private static <T> PlannedDue computePlannedDueGeneric(
+      T followUp,
+      T due,
+      Function<T, String> formatter,
```
bruderj15: try an interface here as well

kadai-adapter/src/main/java/io/kadai/adapter/util/DateTimeUtils.java
```diff
@@ -0,0 +1,82 @@
...
+    public final String due;
+    public final boolean bothSet;
+
+    public PlannedDue(String planned, String due, boolean bothSet) {
+      this.planned = planned;
+      this.due = due;
+      this.bothSet = bothSet;
+    }
+  }
+}
```
bruderj15: Make it a record, omit `bothSet` as arg and infer it instead.

bruderj15 (commented):
Right direction.

The current design still has a flaw though as `computePlannedDueGeneric(..)` contains only a part of the logic.
The invoker(s) currently check(s) `bothSet && enforceServiceLevel => err`.
This, however, is part of the business-logic and should therefore reside inside `computePlannedDueGeneric`.

If you're of opposing view, I am happy to discuss!

leantios:
Would it be possible to perform the Checks in KadaiTaskStarterService instead of in every system connector separately?

bruderj15:
> Would it be possible to perform the Checks in KadaiTaskStarterService instead of in every system connector separately?

Good point! This is where it actually belongs.
However, requires some more abstraction over what a task is I assume. Probably a better fit to think about this with #211.

leantios:
Additionaly, the main timing logic currently lies in TaskInformationMapper::setTimestampsInKadaiTask. This should be checked with the table shown in the issue. On first glance, it seems planned is not set to due -service level if planned is not given but due is

bruderj15:
> Additionaly, the main timing logic currently lies in TaskInformationMapper::setTimestampsInKadaiTask. This should be checked with the table shown in the issue. On first glance, it seems planned is not set to due -service level if planned is not given but due is

@jamesrdi you checked that part?

jamesrdi:
> > Additionaly, the main timing logic currently lies in TaskInformationMapper::setTimestampsInKadaiTask. This should be checked with the table shown in the issue. On first glance, it seems planned is not set to due -service level if planned is not given but due is
> 
> @jamesrdi you checked that part?

The logic for setting planned = due - serviceLevel is done on Kadai, not in KadaiAdapter, you can see in ServiceLevelHandler under the function updatePlannedDueOnCreationOfNewTask() in Kadai

kadai-adapter-camunda7-parent/kadai-adapter-camunda-listener/src/main/java/io/kadai/adapter/camunda/tasklistener/KadaiTaskListener.java
```diff
@@ -228,12 +229,23 @@ private String getReferencedTaskJson(DelegateTask delegateTask) throws JsonProce
     ReferencedTask referencedTask = new ReferencedTask();
 
     referencedTask.setId(delegateTask.getId());
-    referencedTask.setCreated(formatDate(delegateTask.getCreateTime()));
+    referencedTask.setCreated(new DateFormatter().format(delegateTask.getCreateTime()));
```
bruderj15: this allocates every time, consider binding the `DateFormatter` to this class

kadai-adapter-camunda7-parent/kadai-adapter-camunda-listener/src/main/java/io/kadai/adapter/camunda/tasklistener/KadaiTaskListener.java
```diff
@@ -228,12 +229,23 @@ private String getReferencedTaskJson(DelegateTask delegateTask) throws JsonProce
...
     referencedTask.setName(delegateTask.getName());
     referencedTask.setAssignee(delegateTask.getAssignee());
-    referencedTask.setPlanned(formatDate(delegateTask.getFollowUpDate()));
-    referencedTask.setDue(formatDate(delegateTask.getDueDate()));
+
+    Date followUpDate = delegateTask.getFollowUpDate();
+    Date dueDate = delegateTask.getDueDate();
+
+    PlannedDue pd =
+        new PlannedDueComputerDate()
```
bruderj15: same here

kadai-adapter-camunda7-parent/kadai-adapter-camunda-listener/src/test/java/io/kadai/adapter/camunda/tasklistener/KadaiTaskListenerLogicTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+  private final ObjectMapper objectMapper = JacksonConfigurator.createAndConfigureObjectMapper();
+
+  @MethodSource("provideListenerCases")
+  @ParameterizedTest
+  void should_HandlePlannedAndDueCases(
+      Date followUpDate,
+      Date dueDate,
+      boolean enforceServiceLevelValidation,
+      boolean expectException)
```
bruderj15: Can we use the actually expected exception, or at least it's class here?
Use `Optional<..>`.

kadai-adapter-camunda7-parent/kadai-adapter-camunda-listener/src/test/java/io/kadai/adapter/camunda/tasklistener/KadaiTaskListenerLogicTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+      if (expectException) {
+        assertThatThrownBy(() -> invokeGetReferencedTaskJson(listener, delegateTask))
+            .isInstanceOf(SystemException.class)
+            .hasMessageContaining("Both followUp and due dates are set")
+            .hasMessageContaining("kadai.servicelevel.validation.enforce");
+        return;
+      }
+
+      boolean expectPlannedNotNull = !(followUpDate == null && dueDate != null);
```
bruderj15: This whole test is super hard to read. Flatten the logic inside and move expected values into the args for the parameterized test.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -24,6 +27,10 @@ public class ReferencedTaskCreator {
 
   private final Camunda8System camunda8System;
 
+  // read property directly instead of injecting CamundaListenerConfiguration
```
bruderj15: vibe-code piece of evidence

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreatorTest.java
```diff
@@ -19,6 +35,52 @@ void should_SplitVariableNames(String concatenatedVariableNames, List<String> ex
...
+  @ParameterizedTest
+  void should_HandlePlannedAndDueNonExceptionCases(OffsetDateTime followUp, OffsetDateTime due) {
+    Camunda8System camunda8System = createMockCamunda8System();
+    ReferencedTaskCreator creator = new ReferencedTaskCreator(camunda8System);
+    ReflectionTestUtils.setField(creator, "enforceServiceLevelValidation", false);
+
+    ActivatedJob job = createMockActivatedJob(followUp, due);
+    ReferencedTask referencedTask = creator.createReferencedTaskFromJob(job);
+
+    if (followUp == null && due == null) {
```
bruderj15: same here, move logic into concrete args

kadai-adapter/src/main/java/io/kadai/adapter/model/PlannedDue.java
```diff
@@ -0,0 +1,6 @@
+package io.kadai.adapter.model;
+
+/**
+ * DTO representing planned/due result from the domain computation.
```
bruderj15: Not a DTO

kadai-adapter/src/main/java/io/kadai/adapter/util/AbstractPlannedDueComputer.java
```diff
@@ -0,0 +1,42 @@
+package io.kadai.adapter.util;
+
+import io.kadai.adapter.model.PlannedDue;
+import io.kadai.common.api.exceptions.SystemException;
+
+/** Generic base class that centralises the planned/due decision logic. */
+public abstract class AbstractPlannedDueComputer<T> implements PlannedDueComputer<T> {
```
bruderj15, line 12: Document `<T>`

kadai-adapter-kadai-connector/src/main/java/io/kadai/adapter/kadaiconnector/api/impl/TaskInformationMapper.java
```diff
@@ -208,12 +208,17 @@ private void setTimestampsInKadaiTask(TaskImpl kadaiTask, ReferencedTask camunda
 
     String due = camundaTask.getDue();
     if (due == null || due.isEmpty() || "null".equals(due)) {
```
leantios: Do we really need this check here if it is then later checked in convertStringToInstant?

jamesrdi: You are right we do not need the check here

kadai-adapter-camunda7-parent/kadai-adapter-camunda-listener/src/test/java/io/kadai/adapter/camunda/tasklistener/KadaiTaskListenerLogicTest.java
```diff
@@ -0,0 +1,175 @@
...
+      throw new IllegalArgumentException("Cannot parse temporal string: " + s, e);
+    }
+  }
+
+  private static void assertInstantsClose(Instant expected, Instant actual) {
+    long diffMillis = Math.abs(java.time.Duration.between(expected, actual).toMillis());
+    assertThat(diffMillis)
+        .withFailMessage(
+            "Expected instants to be within 6000ms but difference was %d ms", diffMillis)
+        .isLessThanOrEqualTo(6000);
```
jamesrdi, line 134: I had to change this to 6 seconds after defining the expected planned in the arguments for the case planned and due are null

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -35,15 +41,25 @@ public ReferencedTaskCreator(Camunda8System camunda8System) {
    *
    * @param job The ActivatedJob containing the task information.
    * @return A ReferencedTask object populated with data from the job.
+   * @throws SystemException if both follow-up and due dates are set and service level validation is
+   *     enforced.
```
bruderj15, line 45: This is misleading, isn't it?
If both are set and service-level validation is enforced we only throw an exception if service-level does not exactly match difference of both, right?

Therefore docs should be adjusted accordingly.
Please also verify again, that the impl does exactly that and not blindly throws? (I am not 100% certain here with the business-requirement).
Also check that we have test-cases covering exactly this.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreatorTest.java
```diff
@@ -19,6 +35,49 @@ void should_SplitVariableNames(String concatenatedVariableNames, List<String> ex
...
+
+    if (expectedDueInstant != null) {
+      assertInstantsClose(expectedDueInstant, toInstant(referencedTask.getDue()));
+    } else {
+      assertThat(referencedTask.getDue()).isNull();
+    }
+  }
+
+  @Test
+  void should_ThrowSystemException_When_BothAreSetAndEnforcementIsTrue() {
```
bruderj15, line 66: as mentioned above i think this needs differentiation

kadai-adapter/src/main/java/io/kadai/adapter/model/PlannedDue.java
```diff
@@ -0,0 +1,6 @@
+package io.kadai.adapter.model;
+
+/**
+ * Record representing planned/due result from the domain computation.
+ */
+public record PlannedDue(String planned, String due) {}
```
bruderj15, line 6: Any type residing in the domain should have rich fields. I'd much prefer some date-time type instead of a string!

kadai-adapter/src/main/java/io/kadai/adapter/util/AbstractPlannedDueComputer.java
```diff
@@ -0,0 +1,47 @@
...
+  public PlannedDue computePlannedDue(T followUp, T due, boolean enforce) {
+    boolean followUpSet = followUp != null;
+    boolean dueSet = due != null;
+
+    String followUpFormatted = followUpSet ? formatter.format(followUp) : null;
+    String dueFormatted = dueSet ? formatter.format(due) : null;
+    String nowFormatted = formatter.format(now());
+
+    if (followUpSet && dueSet && enforce) {
+      throw new SystemException(
```
bruderj15, line 32: as mentioned twice above

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
jamesrdi: From what I understand from #239 , whenever followUp and Due are set, and enforce is true, then we want to throw error, or do I understand it wrong?

kadai-adapter/src/main/java/io/kadai/adapter/model/PlannedDue.java
jamesrdi: Camunda 7 and Camunda 8 use different type of date type, Camunda 7 uses java.util.date whereas Camunda 8 uses java.time.OffsetDateTime. If we want to use the date-time type, then we have to make 2 records? And in the end we want to set the date in string format for the referencedTask anyways. What do you think?

jamesrdi: I think I can make the type to be generic, if that is better.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
bruderj15: That's certainly what the table says. However form a business-requirement POV this makes no sense.
If both as well as enforce-flag are set, then one must differentiate between the enforced level matching the diff of both or not:
- If matches: Fine, both values are set already - nothing to change
- If not matches: Error

I will bring this up in an internal discussion but am pretty sure we'll go with my proposal.

kadai-adapter/src/main/java/io/kadai/adapter/model/PlannedDue.java
bruderj15: Use `OffsetDateTime` then as it's a superset of `Date`.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
jamesrdi: Well in that case, Kadai already has that logic where it verifies if the planned and due match the service level if enforceServiceLevel is true, so in KadaiAdapter I will just then pass the planned and due as it is in the referenced task without checking if they match the service level, and let kadai check it. You can see it under `updatePlannedDueOnCreationOfNewTask()` in ServiceLevelHandler in Kadai

bruderj15: I don't like this as it leaves the implementation to compute `PlannedDue` brittle in the Adapter.
Handle it entirely in here.

jamesrdi:
As discussed, we want to just implement this logic in kadai. Therefore I will close this PR.

PR #364, opened 2026-01-08
Closes #221 - Fix cycle where C8-Tasks would be attempted to be cancelled multiple times

### LINKED ISSUE #221: C8: unwanted reaction when completing a task in kadai
- when a task is completed in kadai camunda is informed by the REST API 
- then camunda informs kadai again about the completion by the user-task-listener
- we need to make sure that kadai know that the task has already been completed to supress unnecessary operations

C7:
- in the method `setCompletionByKadaiAdapterAsLocalVariable` we set a local camunda variable when informing camunda about the completion 
- then in `taskWasCompletedByKadaiAdapter` we check for this variable and ignore the listener notification if the task comes originaly from kadai

bruderj15:
## Definition of Done

    - will be squashed on merge
  - While this technically is a breaking change due to the external task-id's pattern being changed (`c8sysid-<camunda8-system-id>-utk-<user-task-key>-eik-<element-instance-key>`), we won't treat it as one as the C8-Plugin was experimental until now


kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/Camunda8TaskCompleter.java
```diff
@@ -41,9 +42,12 @@ public SystemResponse completeCamunda8Task(ReferencedTask referencedTask) {
 
     if (completingEnabled) {
       try {
+        final String variables =
+            String.format(
+                "{%s, %s}", referencedTask.getVariables(), TASK_COMPLETED_BY_KADAI_KV_PAIR);
         camundaClient
             .newCompleteUserTaskCommand(getUserTaskKeyFromReferencedTask(referencedTask))
-            .variables("{" + referencedTask.getVariables() + "}")
+            .variables(variables)
```
leantios: How about using [this endpt](https://docs.camunda.io/docs/apis-tools/orchestration-cluster-api-rest/specifications/create-element-instance-variables/): 
```
camundaClient.newSetVariablesCommand(elementInstanceKey)
            .variable(variables)
            .local(true)
            .send()
            .join();
```
The only downside is, that we need the elementInstanceKey here.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/api/impl/Camunda8TaskCompleterTest.java
```diff
@@ -64,5 +64,46 @@ void should_CompleteCamundaTask_When_KadaiTaskIsCompleted() throws Exception {
...
+
+    final Task completedKadaiTask = kadaiEngine.getTaskService().getTask(kadaiTask.getId());
+    assertThat(completedKadaiTask.getState()).isEqualTo(TaskState.CANCELLED);
+    String externalId = kadaiTask.getExternalId();
+
+    long camundaTaskKey = Long.parseLong(externalId.substring(externalId.lastIndexOf("-") + 1));
+    camunda8TestUtil.waitUntil(
+        () -> "COMPLETED".equals(camunda8TestUtil.getCamundaTaskStatus(camundaTaskKey)));
+
+    CamundaAssert.assertThat(processInstance).isCompleted();
```
leantios, line 162: Can we also check the monitoredRun here?

leantios (changes requested):

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/api/impl/Camunda8TaskCompleter.java
bruderj15: Why though? It already works with completion.

Edit: Ah the local part.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/api/impl/Camunda8TaskCompleterTest.java
bruderj15: Technically yes - with some tricks.
Semantically it doesn't make sense here though.
I don't want to add it.

bruderj15:
Also note #371

bruderj15:
I recommend going with this - proven to be working - approach for now.
In parallel extend Kadai with Force-Cancellation in the Java Api (https://github.com/kadai-io/kadai/issues/1641).
If that works as expected and we get it done and integrated in time, **prefer the force-cancellation**  for release.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCompletion.java
```diff
@@ -34,11 +36,20 @@ public UserTaskCompletion(
   @JobWorker(type = "kadai-receive-task-completed-event")
   public void receiveTaskCompletedEvent(final ActivatedJob job)
       throws TaskTerminationFailedException {
+    System.out.println("Foo:" + job.getVariables());
```
leantios: Whoopsie

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
```diff
@@ -76,6 +76,42 @@ public ReferencedTask createReferencedTaskFromJob(ActivatedJob job) {
...
 
+  public static String resolveTaskId(ActivatedJob activatedJob, Camunda8System camunda8System) {
+    return String.format(
+        "c8sysid-%d-utk-%d-eik-%d",
+        camunda8System.getIdentifier(),
+        activatedJob.getUserTask().getUserTaskKey(),
+        activatedJob.getElementInstanceKey());
+  }
+
+  public static Long extractUserTaskKeyFromTaskId(String taskId) {
```
leantios, line 87: Feels wrong to have this here due to Single Responsibility Principle: this is a class for creating a referenced task, not a class with all kinds of utilities

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCancellation.java
```diff
@@ -1,5 +1,7 @@
 package io.kadai.adapter.systemconnector.camunda.tasklistener;
 
+import static io.kadai.adapter.systemconnector.camunda.api.impl.Camunda8TaskCompleter.TASK_COMPLETED_BY_KADAI_KV_PAIR;
```
leantios: I am not completely happy about this solution. Maybe define a class with Constants and add it there?

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/java/io/kadai/adapter/systemconnector/camunda/api/impl/Camunda8TaskCompleterTest.java
```diff
@@ -61,8 +62,49 @@ void should_CompleteCamundaTask_When_KadaiTaskIsCompleted() throws Exception {
...
+
+    final Task completedKadaiTask = kadaiEngine.getTaskService().getTask(kadaiTask.getId());
+    assertThat(completedKadaiTask.getState()).isEqualTo(TaskState.CANCELLED);
+    String externalId = kadaiTask.getExternalId();
+
+    long camundaTaskKey = ReferencedTaskCreator.extractUserTaskKeyFromTaskId(externalId);
+    camunda8TestUtil.waitUntil(
+        () -> "COMPLETED".equals(camunda8TestUtil.getCamundaTaskStatus(camundaTaskKey)));
+
+    CamundaAssert.assertThat(processInstance).isCompleted();
```
leantios, line 162: We should test here whether the variable was set

leantios (changes requested):

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/util/ReferencedTaskCreator.java
bruderj15: Tbf, this isn't perfect, yes. In fact `ReferencedTask` itself should encapsulate this information by exposing a getter.
This however, is caused by #211.

What I did here is then better than scattering the code-base which would be the result of your proposal.
In here, `ReferencedTaskCreator` is the owner of the logic for creation of discussed field.
Hence why it is this class that must expose the information and therefore hide how to decompose the string.

kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/main/java/io/kadai/adapter/systemconnector/camunda/tasklistener/UserTaskCancellation.java
bruderj15: Similar argument as above. It is `Camunda8TaskCompleter` owning the behavior. Therefore the constant should be defined there.

bruderj15: You were right that we would decouple completer and task-listener by introducing a central (bus-like) place.
Overkill here though.

bruderj15:
I kept the composite key although we don't need the element-instance-key anymore.
Imo good to be in there.

leantios (dismissed):
Looks good to me :)

PR #367, opened 2026-01-14
Closes #366 - Bump spring boot to 3.5.9

### LINKED ISSUE #366: Bump Spring Boot to 3.5.9
Dependabot missed out. Probably due to #311

bruderj15:
Aligned HTTP-Client usage to introduced deprecations.

## Definition of Done

    - will be squashed on merge


grittaweisheit (approved):
If we need  the setup logic for a httpClient with props somewhere else in the future we could think about moving the logic to one single point.

PR #370, opened 2026-01-19
Closes #369 - Camunda 8 to use Process-Instance-Key for Business-Process-Id

### LINKED ISSUE #369: Use Process Instance Key for Business Process Id
Like in C7, the Business Process ID should be the key of the process instance and not the key of the bpmn process.

bruderj15:
## Definition of Done

    - will be squashed on merge


leantios:
We don't have any Tests where we verify this behavior?

CRoberto1926 (commented):
LGTM, but I let @leantios approve :)

leantios (approved):
Thanks for the test 😄

PR #384, opened 2026-02-03
Closes #368 - Fix Camunda 8 JSON-Variables Parse-Error for String

### LINKED ISSUE #368: Camunda 8 JSON for variables isn't properly parsed leading to JSON type-mismatches downstream

bruderj15:
The problem was in mapping Camunda-Process-Variables to Kadai custom-attributes.
Strings were not properly enclosed with extra quotes and therefore could not be parsed back into JSON downstream.

The code fixed here is shared between all plugins as it's part of the `kadai-connector`.

The problem did not occur for Camunda 7 as JSON-Values are always JSON-Objects there, e.g. custom-attributes looked like this:
```bash
{camunda:attribute2={"valueInfo":null,"type":"integer","value":5}, camunda:attribute1={"valueInfo":{"objectTypeName":"io.kadai.impl.ProcessVariableTestObject","serializationDataFormat":"application/json"},"type":"object","value":"{\"stringField\":\"\\fForm feed \\b Backspace \\t Tab \\\\Backslash \\n newLine \\r Carriage return \\\" DoubleQuote\",\"intField\":1,\"doubleField\":1.1,\"booleanField\":false,\"processVariableTestObjectTwoField\":[{\"stringFieldObjectTwo\":\"stringValueObjectTwo\",\"intFieldObjectTwo\":2,\"doubleFieldObjectTwo\":2.2,\"booleanFieldObjectTwo\":true,\"dateFieldObjectTwo\":\"1970-01-01 13:12:11\"}]}"}, camunda:attribute3={"valueInfo":null,"type":"boolean","value":true}}
```

In Camunda 8 these are plain JSON and therefore flat. That lead to previous custom-attributes:
```bash
{camunda:attribute1="42", camunda:attribute2="foo"}
```

After applying the fix this leads to:
```bash
{camunda:attribute1="42", camunda:attribute2="\"foo\""}
```

## Definition of Done

    - will be squashed on merge

### DIFF
--- a/kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/resources/processes/sayHello.bpmn
+++ b/kadai-adapter-camunda8-parent/kadai-adapter-camunda-8-system-connector/src/test/resources/processes/sayHello.bpmn
@@ -25,7 +25,7 @@
           <zeebe:input source="=&#34;L11010&#34;" target="kadai_classification_key" />
           <zeebe:input source="=&#34;DOMAIN_A&#34;" target="kadai_domain" />
           <zeebe:input source="=&#34;attr1,attr2&#34;" target="kadai_attributes" />
-          <zeebe:input source="=&#34;1&#34;" target="attr1" />
+          <zeebe:input source="=&#34;foo&#34;" target="attr1" />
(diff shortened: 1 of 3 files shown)


grittaweisheit (approved):
Looks good!

leantios (approved):

PR #437, opened 2026-04-02
Closes #430 - Multi-Thread `KadaiTaskStarterOrchestrator`

### LINKED ISSUE #430: Parallelize `KadaiTaskStarterOrchestrator` to improve resource-utilization
We already declare parallelized `@Scheduler`s via `SchedulerConfig`.
Verify that this doesn't act on single scheduler instances but globally (I think so - if it wasn't, we would already have parallelized).

Then parallelize the `KadaiTaskStarterOrchestrator`.
Also consider parallelizing other similar components.

Ideally verify parallelization via a test.

bruderj15:
## Definition of Done

    - will be squashed on merge
- [x] If the documentation needs an update, following there's a link to it: https://github.com/kadai-io/kadai-doc/pull/217

### DIFF
--- a/kadai-adapter/src/main/java/io/kadai/adapter/impl/scheduled/KadaiTaskStarterOrchestrator.java
+++ b/kadai-adapter/src/main/java/io/kadai/adapter/impl/scheduled/KadaiTaskStarterOrchestrator.java
@@ -31,6 +31,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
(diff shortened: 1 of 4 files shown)


kadai-adapter/src/main/java/io/kadai/adapter/impl/scheduled/KadaiTaskStarterOrchestrator.java
```diff
@@ -124,32 +133,52 @@ public Duration getExpectedRunDuration() {
...
-              "caught Exception when attempting to start KadaiTask for referencedTask {}");
-        }
-      } catch (Exception e) {
-        handleError(
-            systemConnector,
-            referencedTask,
-            e,
-            "caught unexpected Exception when attempting to start KadaiTask "
-                + "for referencedTask {}");
+        future.get();
```
CRoberto1926, line 172: If I understood correctly:
- all the `tasksToStart` are submitted for parallel creation and start
- here we iterate on the `futures` and in the case that one `future` has not completed, we block the main thread and wait for it

Is this what we want?

bruderj15: > all the tasksToStart are submitted for parallel creation and start

yeah, for `n` tasks `min(n, threadConfig)`

> here we iterate on the futures and in the case that one future has not completed, we block the main thread and wait for 

Yeah. While we wait for any single one, the others are still running. We have to wait for the slowest one no matter what.

> Is this what we want?

At least what I recommend. Is there a better approach?

CRoberto1926 (approved):

kadai-adapter/src/main/java/io/kadai/adapter/impl/scheduled/KadaiTaskStarterOrchestrator.java
CRoberto1926: if we have to wait anyway for the slowest, probably not

PR #468, opened 2026-06-19
Closes #420 - Fix enforcement of Transaction-Boundaries on Cluster-Locking

### LINKED ISSUE #420: Task-Lock for clustered environments may not be applied correctly
_From Prod:_

When multiple Adapters are running in a cluster, frequently below error pops up:

```json
{
  "@timestamp": "2026-03-05T22:48:09.370+01:00",
  "level": "ERROR",
  "class":" i.k.a.k.a.impl.KadaiSystemConnectorImpl",
  "message": "Caught Exception TaskAlreadyExistException [errorCode=ErrorCode [key=TASK_ALREADY_EXISTS,…"
}
```

All tasks are successfully created in the end. However the lock on those tasks causing an errors doesn't seem to be applied or respected correctly.

bruderj15:
## Definition of Done

    - will be squashed on merge


kadai-adapter-camunda7-parent/kadai-adapter-camunda-outbox-rest/src/main/java/io/kadai/adapter/camunda/outbox/rest/service/Camunda7TaskEventsService.java
```diff
@@ -591,9 +594,19 @@ private List<Camunda7TaskEvent> getCompleteAndDeleteEvents(Duration lockDuration
   }
 
   private Connection getConnection() {
```
leantios, line 612: Maybe rename to `getConnectionWithAutoCommit` for clarity?

kadai-adapter-camunda7-parent/kadai-adapter-camunda-outbox-rest/src/main/java/io/kadai/adapter/camunda/outbox/rest/service/Camunda7TaskEventsService.java
```diff
@@ -469,6 +470,7 @@ private List<Camunda7TaskEvent> getCreateEvents(Duration lockDuration) {
         if (ids != null) {
           try {
             unlockEvents(ids, connection);
```
leantios, line 482: Just out of curiosity: why do we unlock here right after locking without any operation in between? Why is the locking then even necessary?

leantios (commented):
Looks very good! Just a general question: don't we need `connection.commit();` whenever we lock Events? Why do we only need it for `getCreateEvents`?

kadai-adapter-camunda7-parent/kadai-adapter-camunda-outbox-rest/src/main/java/io/kadai/adapter/camunda/outbox/rest/service/Camunda7TaskEventsService.java
bruderj15: Very good question.
Seems weird - even wrong - but is correct.

When `Connection::commit` throws an exception, it doesn`t mean that the database-level transaction failed. It _may_ have gone through successfully (think any reason an indeterminate outcome: network failure, ...).
Therefore, unlocking again is the most bulletproof approach.

This is a _little_ more visible now with the explicit transaction-lifecycle but already was valid when auto-commit was in place - though I doubt that the previous implementor had a clue of this.

kadai-adapter-camunda7-parent/kadai-adapter-camunda-outbox-rest/src/main/java/io/kadai/adapter/camunda/outbox/rest/service/Camunda7TaskEventsService.java
bruderj15: overloading totally fine here imo

bruderj15:
@leantios 

> Just a general question: don't we need `connection.commit();` whenever we lock Events? Why do we only need it for `getCreateEvents`?

Yes, it's always required. Auto-Commit does it for us as `getConnection()` without args is just `getConnection(true)`, hence defaults to a connection with auto-commit.
For all usages of `(un)lockEvents` in `getCreateEvents` (which invokes only `getConnection(false)`) it already holds: explicit commit after `(un)lockEvents`

bruderj15:
> Just a general question: don't we need connection.commit(); whenever we lock Events? Why do we only need it for getCreateEvents?

I understand your concern now. 
Yes we must investigate whether the wrong transaction-boundary also affects the other `get...Events` functions.

leantios (approved):
Good job!

